PR #3, opened 2023-06-30
2 khiops coclustering simplification fails on non ascii data

### LINKED ISSUE #2: Khiops coclustering simplification fails on non-ASCII data
#### Description

Currently, when the `simplify` method is called on a `KhiopsCoclustering` estimator instance that was fitted on data having variables whose values contain non-ASCII characters, it fails to decode the estimator instance's raw report correctly.

The fitted estimator instance's raw report JSON is dumped via the `json.dumps` function which by default escapes all non-ASCII characters. This results in a UTF-8 encoded *escaped* Unicode characters, which hampers reading by Khiops (according to the scenario of the `simplify_coclustering` task), which expects an input file with UTF-8-encoded data of characters as they are (with no escaping).

#### Questions/Ideas
- The solution is to set `json.dumps`'s `ensure_ascii` parameter to `False` (it is `True` by default).
   - The JSON report dumping logic based on the above considerations should be encapsulated into a method (named `dump_khiops_json_file`) of the `KhiopsJSONObject` class in the `core.internals.io` module.

popescu-v:
Closes #2.

### DIFF
--- a/.github/workflows/pre-commit-checks.yml
+++ b/.github/workflows/pre-commit-checks.yml
@@ -1,0 +1,25 @@
+name: "Run pre-commit checks"
+
+on:
+  pull_request:
+  push:
+  workflow_dispatch:
+
+jobs:
+  pre-commit-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+
+      - name: Setup Python
+        uses: actions/setup-python@v3
+
+      - name: Install khiops-python dependencies
+        run: |
+          python setup.py egg_info
+          pip install `grep -v "^\[" pykhiops.egg-info/requires.txt`
+          rm -rf pykhiops.egg-info
+
+      - name: Run pre-commit checks
+        uses: pre-commit/action@v3.0.0
--- a/pykhiops/core/internals/io.py
+++ b/pykhiops/core/internals/io.py
@@ -5,6 +5,7 @@
 # see the "LICENSE.md" file for more details.                                        #
 ######################################################################################
 """Classes to handle Khiops specific I/O"""
+import copy
 import io
 import json
 import os
@@ -109,6 +110,7 @@ class KhiopsJSONObject:
         self.khiops_encoding = "utf8"
         self.ansi_chars = []
         self.colliding_utf8_chars = []
+        self.json_data = None
 
         # Initialize from json data
         if json_data is not None:
@@ -165,6 +167,9 @@ class KhiopsJSONObject:
                 self.ansi_chars = json_data["ansi_chars"]
                 if "colliding_utf8_chars" in json_data:
                     self.colliding_utf8_chars = json_data["colliding_utf8_chars"]
+
+            # Store a copy of the data to be able to write copies of it
+            self.json_data = copy.deepcopy(json_data)
 
     def create_output_file_writer(self, stream):
         """Creates an output file with the proper encoding settings
@@ -230,6 +235,23 @@ class KhiopsJSONObject:
             raise PyKhiopsJSONError(
                 f"Could not load Khiops JSON file: {json_file_path}"
             ) from error
+
+    def write_khiops_json_file(self, json_file_path):
+        """Write the JSON data of the object to a Khiops JSON file
+
+        Parameters
+        ----------
+        json_file_path : str
+            Path to the Khiops JSON file.
+        """
+        if self.json_data is not None:
+            # Serialize JSON data to string
+            # Do not escape non-ASCII Unicode characters
+            json_string = json.dumps(self.json_data, ensure_ascii=False)
+            with io.BytesIO() as json_stream:
+                writer = self.create_output_file_writer(json_stream)
+                writer.write(json_string)
+                fs.write(uri_or_path=json_file_path, data=json_stream.getvalue())
 
 
 class PyKhiopsOutputWriter:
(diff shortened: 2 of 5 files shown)


pykhiops/core/internals/io.py
```diff
@@ -231,6 +238,17 @@ def load_khiops_json_file(self, json_file_path):
                 f"Could not load Khiops JSON file: {json_file_path}"
             ) from error
 
+    def dump_khiops_json_data(self, json_file_path):
+        """ """
```
felipe-olmos-orange, line 242: rename to `write_khiops_json_file`.

And write the docstring.

pykhiops/core/internals/io.py
```diff
@@ -231,6 +238,17 @@ def load_khiops_json_file(self, json_file_path):
...
+        if self.json_data is not None:
+            # Serialize JSON data to string
+            # Do not escape non-ASCII Unicode characters; keep them as such
+            json_payload = json.dumps(self.json_data, ensure_ascii=False)
```
felipe-olmos-orange, line 246: rename to `json_string`

pykhiops/core/internals/io.py
```diff
@@ -231,6 +238,17 @@ def load_khiops_json_file(self, json_file_path):
                 f"Could not load Khiops JSON file: {json_file_path}"
             ) from error
 
+    def dump_khiops_json_data(self, json_file_path):
+        """ """
+        if self.json_data is not None:
+            # Serialize JSON data to string
+            # Do not escape non-ASCII Unicode characters; keep them as such
```
felipe-olmos-orange, line 245: "keep them as such" is redundant

pykhiops/core/internals/io.py
```diff
@@ -166,6 +168,11 @@ def __init__(self, json_data=None):
                 if "colliding_utf8_chars" in json_data:
                     self.colliding_utf8_chars = json_data["colliding_utf8_chars"]
 
+            # Store actual data, if it satisfies the required constraints
+            # Deep-copy it so that the KhiopsJSONObject's state stays constant
+            # when its init arguments are changed after initialization
```
felipe-olmos-orange, line 173: Just say `Store a copy of the data to be able to write copies of it`

pykhiops/sklearn/estimators.py
```diff
@@ -1012,10 +1012,7 @@ def _simplify(
         full_coclustering_file_path = fs.get_child_path(
             output_dir, "FullCoclustering.khcj"
         )
-        fs.write(
-            full_coclustering_file_path,
-            json.dumps(self.model_report_raw_).encode("utf-8"),
-        )
+        self.model_report_.dump_khiops_json_data(full_coclustering_file_path)
```
felipe-olmos-orange, line 1015: We should deprecate `model_report_raw_` since now can be accessed with `model_report_.json_data`.

pykhiops/core/internals/io.py
```diff
@@ -5,6 +5,7 @@
 # see the "LICENSE.md" file for more details.                                        #
 ######################################################################################
 """Classes to handle Khiops specific I/O"""
+import copy
```
felipe-olmos-orange, line 8: General comment: Before merging add a GitHub workflow to execute the pre-commit checks 
(you can copy that of khiops: https://github.com/KhiopsML/khiops/blob/dev/.github/workflows/pre-commit-checks.yml)

felipe-olmos-orange (changes requested):

pykhiops/core/internals/io.py
popescu-v: Ok; I will do this in the same PR.

pykhiops/sklearn/estimators.py
popescu-v: Ok; I plan to write a property which will raise a deprecation warning on access and return `model_report_.json_data`.

pykhiops/core/internals/io.py
popescu-v: My bad for the docstring. Done.

pykhiops/core/internals/io.py
popescu-v: Done.

pykhiops/core/internals/io.py
popescu-v: Ok; removed.

pykhiops/core/internals/io.py
popescu-v: Ok; rephrased as requested.

pykhiops/sklearn/estimators.py
popescu-v: Done.

pykhiops/core/internals/io.py
popescu-v: Done.

.github/workflows/pre-commit-checks.yml
```diff
@@ -0,0 +1,25 @@
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v3
+
+      - name: Install dependencies
+        run: |
+          python setup.py egg_info \
+          && pip install `grep -v "^\[" pykhiops.egg-info/requires.txt` \
+          && rm -rf pykhiops.egg-info
+
```
felipe-olmos-orange, line 23: This is not necessary

felipe-olmos-orange: Ok, then it is not necessary to do it with `&&`. When you use `|`  it is a if it were a text file.

Also rename to "Install khiops-python dependencies"

popescu-v: Ok, the required changes have been implemented.

felipe-olmos-orange (approved):

PR #95, opened 2023-07-13
Compute versions for "vX.Y.Z" Git tags

### LINKED ISSUE #94: Khiops Version calculation fails on "vx.y.z" tags
#### Description
Git tags are currently of the form "vx.y.z", where "x" and "y" are numbers and "z" starts with a number.
Or, `KhiopsVersion` extracts version from Git tags and expects them to be of the form "x.y.z", with "x", "y", and "z" as above.

#### Questions/Ideas
- We need to remove the "v" prefix from the Git tag string before computing the Khiops version out of it.
  - As Python >= 3.8 is supported, `str.removeprefix` (introduced in Python 3.9) cannot be used. Hence, the removal needs to be programmed using Python <= 3.8 features.

popescu-v:
closes #94

### DIFF
--- a/pykhiops/core/internals/version.py
+++ b/pykhiops/core/internals/version.py
@@ -6,12 +6,28 @@
 ######################################################################################
 """Class to handle Khiops version strings"""
 
+import sys
+
 
 class KhiopsVersion:
     """Encapsulates the Khiops version string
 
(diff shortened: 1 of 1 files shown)


felipe-olmos-orange (approved):
LGTM

PR #116, opened 2023-11-23
Integrate the remote access service into the khiopspydev Docker image

### LINKED ISSUE #114: Enable running the Docker runner remote access tests in the CI
#### Description
Currently, the Docker runner tests are conditioned on setting up a Khiops remote access service in a Docker container. We need to set-up this environment in the GitHub CI context.

#### Questions/Ideas
- A way of doing this would be to make available to the GitHub CI a Docker image package containing Khiops and the remote access service
  - Concretely, we would need to:
    - get / have access to a Docker image that contains Khiops plus the remote access service
    - build a test Docker image that inherits from the Docker image mentioned above, and adds `khiops-python`'s dependencies that are needed for running the tests
    - define three environment variables in the GitHub CI: `KHIOPS_RUNNER_SERVICE_PATH`, `KHIOPS_DOCKER_RUNNER_URL`, and `KHIOPS_DOCKER_RUNNER_SHARED_DIR`, according to what the test Docker image contains
    - launch the Docker runner tests on the test Docker image
- We need to make sure that the Docker image that the test Docker image inherits from contains an up-to-date version of Khiops and of the remote access service that is compatible with Khiops
- We also need to see how the synchronization between the Docker image with the remote access service and Khiops + `khiops-python`'s dependencies can be done

popescu-v:
closes #114

### DIFF
--- a/.github/workflows/build-dev-container.yml
+++ b/.github/workflows/build-dev-container.yml
@@ -3,6 +3,7 @@ name: Build development container
 env:
   DEFAULT_KHIOPS_REVISION: dev
   DEFAULT_KHIOPSDEV_OS: ubuntu22.04
+  DEFAULT_SERVER_REVISION: latest
 on:
   workflow_dispatch:
     inputs:
@@ -12,6 +13,9 @@ on:
       khiopsdev-os:
         default: ubuntu22.04
         description: Github runner OS
+      server-revision:
+        default: latest
+        description: Tag of the server Docker image
   workflow_call:
     inputs:
       khiops-revision:
@@ -22,6 +26,10 @@ on:
         type: string
         default: ubuntu22.04
         description: Github runner OS
+      server-revision:
+        type: string
+        default: latest
+        description: Tag of the server Docker image
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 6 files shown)


tests/test_remote_access.py
```diff
@@ -110,7 +110,7 @@ def test_khiops_classifier_with_remote_access(self):
             """Test the training of a khiops_classifier with remote resources"""
             # Setup paths
             output_dir = (
-                kh.get_runner().khiops_tmp_dir
+                kh.get_runner().khiops_temp_dir
```
folmos-at-orange, line 113: This was a bug ?

.github/workflows/build-dev-container.yml
```diff
@@ -3,6 +3,7 @@ name: Build development container
 env:
```
folmos-at-orange, line 3: **General Comment**: 

So the docker tests run locally ?

.github/workflows/run-unit-tests.yml
```diff
@@ -70,6 +70,9 @@ jobs:
       - name: Run Unit Tests
         env:
           KHIOPS_SAMPLES_DIR: ${{ github.workspace }}/khiops-samples
+          KHIOPS_DOCKER_RUNNER_URL: https://localhost:11000
+          KHIOPS_DOCKER_RUNNER_SHARED_DIR: /tmp/sandbox
+          KHIOPS_RUNNER_SERVICE_PATH: /scripts/run_service.sh
         run: |
           # This is needed so that the Git tag is parsed and the khiops-python
           # version is  # retrieved
```
folmos-at-orange, line 78: Could you please remove the extra `#`?

folmos-at-orange (changes requested):

tests/test_remote_access.py
popescu-v: It's `khiops_temp_dir` that is part of the runner API. `KHIOPS_TMP_DIR` is the name of the relevant environment variable though.

I think the remote runner test code used to work anyway because of duck typing, as the `khiops_tmp_dir` was set in the `setUpClass` methods and used throughout the tests afterwards (by setting the `results_dir` to a path containing `khiops_tmp_dir`).

.github/workflows/run-unit-tests.yml
popescu-v: Ok.

tests/test_remote_access.py
folmos-at-orange: So it was a bug.

.github/workflows/build-dev-container.yml
folmos-at-orange: Ok!

folmos-at-orange (changes requested):

tests/test_remote_access.py
popescu-v: Yes, I would say so. Thus, can this thread be closed?

.github/workflows/build-dev-container.yml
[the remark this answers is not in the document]
popescu-v: Yes, ok. And the same goes for `setup-buildx-action` and `login-action`; I would bump them all to `@v3`.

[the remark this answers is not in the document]
popescu-v: Done.

PR #125, opened 2023-12-08
Pin mpich version on Linux

sgouache:
In an attempt to mitigate the  [bug](https://github.com/KhiopsML/khiops/issues/104) encountered on Linux with many orphan records, pin the mpich version embedded in conda package for ubuntu and debian, python version <= 3.10


popescu-v (commented):
Technically, the changes seem OK to me, as a temporary mitigation of the Khiops bug. However, before approving, we need to settle on the policy to adopt IMHO.

folmos-at-orange:
We will fix the issue from #126 inspired by this PR.

PR #130, opened 2024-01-08
Check existence of docker shared_dir only for local ones

folmos-at-orange:


khiops/extras/docker.py
```diff
@@ -37,7 +38,9 @@ def __init__(self, url, shared_dir, insecure=False):
         self._url = url
         self._insecure = insecure
         self.shared_dir = shared_dir
-        if self.shared_dir is None or not fs.exists(self.shared_dir):
+        if self.shared_dir is None or (
+            fs.is_local_resource(self.shared_dir) and not fs.exists(self.shared_dir)
+        ):
```
popescu-v, line 43: I would separate the two `if` condition parts and:
- would raise `ValueError` if `self.shared_dir is None`
- would raise `RuntimeError` if `self.shared_dir` is a non-existing local directory

popescu-v, line 43: I would also add a test that a `RuntimeError` (see previous comment) is only raised for a non-existing local `shared_dir`.

popescu-v (changes requested):
Almost LGTM, with two caveats (see comments).

khiops/extras/docker.py
folmos-at-orange: Ok

folmos-at-orange: Done

folmos-at-orange: Done

popescu-v: Ok for these checks.

popescu-v: Actually, what I had in mind was an automatic test (in test_remote_access) that a ~`RuntimeError`~ `KhiopsEnvironmentError` is only raised for a non-existing local `shared_dir` (but not for a remote URI).

popescu-v (changes requested):
There is only one comment left to address IMHO.

khiops/extras/docker.py
folmos-at-orange: That is what i did.

folmos-at-orange: Done

tests/test_khiops_integrations.py
```diff
@@ -99,3 +104,13 @@ def test_estimator_multiple_create_and_fit_does_not_raise_exception(self):
         finally:
             if os.path.exists(output_dir):
                 shutil.rmtree(output_dir)
+
+
+class DockerKhiopsEdgeCases(unittest.TestCase):
+    """Test for KhiopsDocker runner edge cases"""
+
+    def test_shared_dir_edge_cases(self):
+        """Test that shared dir exists when it is a local path"""
```
popescu-v, line 113: I would rephrase to something like: "Test that KhiopsEnvironmentError is raised when shared_dir is a local path and does not exist".

khiops/extras/docker.py
```diff
@@ -71,10 +86,12 @@ def _fetch(self, request):
         # HTTPS mode
         response = None
```
popescu-v, line 87: This is not needed anymore IMHO.

khiops/extras/docker.py
```diff
@@ -25,10 +28,30 @@ class KhiopsDockerRunner(KhiopsRunner):
...
 
     Requires a running docker Khiops instance
+
+    Parameters
+    ----------
+
+    url : str
+        URL for the Docker Khiops server.
+    shared_dir : str
+        Path for the shared directory dire. May be an URL.
```
popescu-v, line 38: I would 's/May be an URL/Can be either a local path or an URL/'.

popescu-v, line 38: s/shared directory dire./shared directory./

popescu-v (changes requested):
Almost LGTM, barring 4 remaining comments.

tests/test_khiops_integrations.py
folmos-at-orange: Done

khiops/extras/docker.py
folmos-at-orange: Ok and done

khiops/extras/docker.py
folmos-at-orange: I put "May be an URL/URI" to keep it consistent with the codebase.

popescu-v (approved):
LGTM

PR #135, opened 2024-01-25
Use nested context managers to support Python 3.8

### LINKED ISSUE #134: Syntax error for parenthesized context managers in Python 3.8
#### Description
#132 used parenthesized context managers in the local runner for redirecting the `stdout` and `stderr` of the `module` function to buffers:
```python
with (
    io.StringIO() as stdout_buffer,
    io.StringIO() as stderr_buffer,
    redirect_stdout(stdout_buffer),
    redirect_stderr(stderr_buffer),
):
    ...
```
Or, parenthesized context managers (
```python
with (ctx1 as x1, .. ctx2 as x2):
    ...
```
)
were introduced by PEP 617 https://www.python.org/dev/peps/pep-0617/, which was accepted in Python 3.9 (https://docs.python.org/3/whatsnew/3.9.html#new-parser). Hence this syntax is not supported in Python 3.8.

#### Questions/Ideas
- Hence, in Python 3.8 we need to either:
   - use nested context managers:
     ```python
     with ctx1 as x1:
         with ctx2 as x2:
             ...
     ```, or
   - use line-terminating characters:
      ```python
      with ctx1 as x1, \
          ctx2 as x2:
          ...
      ```

popescu-v:
closes #134

### DIFF
--- a/khiops/core/internals/runner.py
+++ b/khiops/core/internals/runner.py
@@ -1116,15 +1116,16 @@ class KhiopsLocalRunner(KhiopsRunner):
 
                 module = env_modules_api.module
                 mpich_module = None
-                with (
-                    io.StringIO() as stdout_buffer,
-                    io.StringIO() as stderr_buffer,
-                    redirect_stdout(stdout_buffer),
-                    redirect_stderr(stderr_buffer),
-                ):
-                    module_availability_status = module("avail")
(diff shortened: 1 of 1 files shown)


folmos-at-orange (approved):
LGTM

PR #144, opened 2024-02-09
Implement new Khiops versioning scheme

folmos-at-orange:


folmos-at-orange:
In the end we were already compatible with a fourth digit. The thing that needed to be addressed were the new Khiops versioning scheme.

khiops/core/internals/version.py
```diff
@@ -9,149 +9,137 @@
...
-            self._trailing_char is not None
-            and self._trailing_char not in self._allowed_chars
-        ):
-            raise ValueError(
-                f"Trailing char of version string '{version}' "
-                f"must be one of {self._allowed_chars}"
+        raw_parts = re.sub("^v", "", self._version_str).split(".")
+
+        # Check the Khiops version format: MAJOR.MINOR.PATCH[-PRE_RELEASE]
+        if len(raw_parts) != 3:
```
popescu-v, line 36: This implies that all tags on the Khiops GitHub repo will observe this convention, right?

khiops/core/internals/version.py
```diff
@@ -9,149 +9,137 @@
...
+        if _is_simple_digit(patch_and_pre_release):
+            self._patch = int(patch_and_pre_release)
+            self._pre_release = None
+        else:
+            if "-" not in patch_and_pre_release:
+                self._raise_init_error(
+                    "PATCH-PRE-RELEASE version part does not contain '-'",
+                    version_str,
+                )
+            self._patch, self._pre_release = patch_and_pre_release.split("-")
```
popescu-v, line 66: IMHO, we need to check that `patch_and_pre_release` contains a single `-`

khiops/core/internals/version.py
```diff
@@ -9,149 +9,137 @@
...
-            patch_number = self._parts[2]
-        return patch_number
+        return self._patch
+
+    @property
+    def pre_release(self):
+        """str : The pre-release tag of this version. Either 'a', 'b' or 'rc' or None"""
+        return self._pre_release
 
     def __str__(self):
```
popescu-v, line 100: `__str__` could just call `__repr__`, couldn't it?

khiops/core/internals/version.py
```diff
@@ -9,149 +9,137 @@
...
+        )
 
     def __gt__(self, version):
-        if self == version:
-            return False
-        elif self._parts > version._parts:
-            return True
-        elif self._parts < version._parts:
-            return False
+        """Lexicographic order for each of the version parts
```
popescu-v, line 115: The way I read it is rather "Numeric order for major, minor and patch parts. Lexicographic order for the pre-release part", right?

khiops/core/internals/version.py
```diff
@@ -163,4 +151,4 @@ def __le__(self, version):
         return not self.__gt__(version)
 
     def __hash__(self):
-        return self.version.__hash__()
+        return self._version_str.__hash__()
```
popescu-v, line 154: Or `self.__repr__().__hash__()`?

popescu-v (changes requested):
See comments

khiops/core/internals/version.py
```diff
@@ -9,149 +9,137 @@
 import re
 
 
+def _is_simple_digit(string):
+    """Tests if a stringos contains only characters [0-9]
```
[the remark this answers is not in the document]
folmos-at-orange: Ok & Done

khiops/core/internals/version.py
folmos-at-orange: Yes

khiops/core/internals/version.py
folmos-at-orange: Ok

folmos-at-orange: Changed the test just above that line to cover that case. Done.

khiops/core/internals/version.py
folmos-at-orange: Removed the method.

khiops/core/internals/version.py
folmos-at-orange: major, minor and patch follow numerical order and the pre-release follows an alphabetical order. The whole version is lexicographical, taking each part as a "character" (cf. https://en.wikipedia.org/wiki/Lexicographic_order#Definition)

khiops/core/internals/version.py
folmos-at-orange: Ok.

popescu-v:
> In the end we were already compatible with a fourth digit. The thing that needed to be addressed were the new Khiops versioning scheme.

Yes, ok, I got it. So, no need for extra provisions for the fourth digit

khiops/core/internals/version.py
popescu-v: Ok, so the "symbols" are the parts, right? Hence "a.b.c > a'.b'.c'" iff "a > a' or a == a' and b > b' or a == a' and b == b' and c > c'", knowing that "a > a' and "b > b'" are determined numerically, while "c > c'" is determined lexicographically, taking "c" and "c'" characters as "symbols", right?

If all this is true, then I think the comment should be updated: otherwise, IMHO we understand that for "each of the version parts" the ordering relation is lexicographic. Whereas what we have here is lexicographic order in terms of the versions parts, right? Then, I would at least propose something like: "Lexicographic order in terms of the version parts."

khiops/core/internals/version.py
popescu-v: Ok, thus we remain with the default `__str__`, which just calls `__repr__`

khiops/core/internals/version.py
folmos-at-orange: Maybe that's too much. Let's forget lexicographical and just put "Normal versioning order" .

folmos-at-orange: Ok as per offline conversation.

folmos-at-orange: Done

khiops/core/internals/version.py
```diff
@@ -9,149 +9,163 @@
...
-        else:
-            patch_number = self._parts[2]
-        return patch_number
+        return self._patch
 
-    def __str__(self):
-        return self.version
+    @property
+    def pre_release(self):
+        """str : The pre-release tag of this version. Either 'a', 'b' or 'rc' or None"""
```
popescu-v, line 123: Update to mention the increment number.

khiops/core/internals/version.py
```diff
@@ -9,149 +9,163 @@
...
-            this_padded_parts == padded_parts
-            and self._trailing_char == version._trailing_char
-        ):
-            return True
-        return False
+        return (
+            self.major == version.major
+            and self.minor == version.minor
+            and self.patch == version.patch
+            and self.pre_release == version.pre_release
```
popescu-v, line 137: If pre-release increment has several digits, how is the comparison being done? Two choices:
- lexicographically (as of now): "-a2" > "-a12"
- numerically: "-a2" < "-a12"

Only the latter  seems correct to me. Hence, IMHO, this part needs to be updated: the pre-release part needs to be compared lexicographically as before, while the pre-release increment needs to be compared numerically.

khiops/core/internals/version.py
```diff
@@ -9,149 +9,163 @@
...
+        elif self.patch > version.patch:
+            is_greater = True
+        elif self.patch < version.patch:
+            is_greater = False
+        elif self.pre_release is None and version.pre_release is not None:
+            is_greater = True
+        elif self.pre_release is not None and version.pre_release is None:
+            is_greater = False
+        elif self.pre_release is not None and version.pre_release is not None:
+            is_greater = self.pre_release > version.pre_release
```
popescu-v, line 163: Update to separate lexicographic comparison of pre-release tags and numeric comparisons of pre-release increments.

tests/test_core.py
```diff
@@ -2144,24 +2144,25 @@ class KhiopsCoreVariousTests(unittest.TestCase):
...
+            "9.5.1-a1",
+            "9.5.1-a2",
+            "9.5.1-b",
             "9.5.1",
-            "10",
+            "10.0.0",
             "10.0.1",
-            "10.0.8b",
+            "10.0.8-b",
+            "10.0.8-b2",
```
popescu-v, line 2156: Add e.g. "10.0.8-b12" and check proper comparison, i.e. that "10.0.8-b2" < "10.0.8-b12".

popescu-v (changes requested):
A few comments need to be addresssed.

khiops/core/internals/version.py
folmos-at-orange: I'll make the comparison by the internal parts.

tests/test_core.py
folmos-at-orange: Ok

folmos-at-orange: And I won't accept pre-releases without increment.

folmos-at-orange: And also left zeros are forbidden.

popescu-v: Ok and ok.

khiops/core/internals/version.py
folmos-at-orange: Done

khiops/core/internals/version.py
folmos-at-orange: Done, now it is numerically.

khiops/core/internals/version.py
folmos-at-orange: Done

popescu-v (approved):
LGTM.

PR #149, opened 2024-02-15
Release 10.2.0.0

folmos-at-orange:


popescu-v (approved):
- CI/CD workflow logs seem OK.
- Conda package installs OK (channel looks OK).
- Generated documentation looks OK.
- Attention point: currently, last tag (according to the previous version format) is being used  
   for computing the version information for Pip and Conda packages. Need to check 
   whether the new version format works OK in all workflow and for all packages.

The last point notwithstanding, LGTM.

PR #160, opened 2024-03-14
Support single-column Pandas dataframes as targets in estimators

tramora:
the target variable "y" can currently be an array-like of shape (n_samples,), a pandas.Dataframe of shame (n_samples, 1) - purpose of this commit - or a string key of a dataset dict (will be removed in version 11)


khiops/sklearn/estimators.py
```diff
@@ -301,6 +301,12 @@ def fit(self, X, y=None, **kwargs):
         initial_runner_temp_dir = kh.get_runner().root_temp_dir
         kh.get_runner().root_temp_dir = computation_dir
 
+        # if a DataFrame is given, extract the first column to get an array-like object
```
popescu-v, line 304: The place for this kind of processing is in `tables.py`, namely in the `DatasetTable` class and its specializations, viz. `PandasTable` (for the cases where `X` is a dataframe). The case where `X` is a NumPy array is already being taken care of in the `NumpyTable` class.

But, before converting the dataframe to a series, we need to check that, as per the issue's content, the input dataframe `y` is a has precisely one column. In this case, even if the dataframe is empty (has 0 rows), `iloc` does not fail.

However, `column_or_1d` is not fully equivalent to `iloc`, as `iloc` preserves the Series type and the name, whereas `column_or_1d` itself does not. To be fully equivalent to `foo.iloc[:, 0]`, we need to do: `pd.Series(column_or_1d(foo), name=foo.columns[0])` (under the hypothesis that the dataframe `foo` has one column).

Hence, your proposed modification could be moved to `tables.py`, in the `PandasTable.__init__(...)` method, where the `target_column` argument is being type-checked and value-checked. More precisely, after the `if isinstance(target_column, pd.Series): ...` block, I would add the following block:

```python
            ...
            elif isinstance(target_column, pd.DataFrame):
                number_of_target_columns = len(target_column.columns)
                if number_of_target_columns != 1:
                    raise ValueError(
                        "Target dataframe should contain only one column. "
                        f"It contains {number_of_target_columns}."
                    )
                target_column = target_column.iloc[:, 0]
                # Or:
                # target_column = pd.Series(
                #     column_or_1d(target_column), name=target_column.columns[0]
                # )
            ...
```

tests/test_sklearn.py
```diff
@@ -1779,6 +1780,10 @@ def _test_template(
                 custom_kwargs.get("fit", {}) if custom_kwargs is not None else {}
             )
             estimator.fit(X_train_data, y_train_data, **fit_kwargs)
+            # FIXME : find a clever way to avoid executing the estimator method twice
```
popescu-v, line 1783: Here, we need to create a brand new test. We should not change `_test_template` in `test_sklearn.py`, but add a new test. To this end, we need to add a new input to the sklearn tests fixture.

popescu-v (changes requested):
There are two pending comments to address.

khiops/sklearn/estimators.py
tramora: Indeed `tables.py` was the right place for the modification.

tests/test_sklearn.py
```diff
@@ -2021,6 +2078,17 @@ def test_parameter_transfer_regressor_fit_from_monotable_dataframe(self):
             source_type="dataframe",
         )
 
+    def test_parameter_transfer_regressor_fit_from_monotable_dataframe_with_df_y(
+        self,
+    ):
+        """Test parameter transfer from monotable dataframe fit to core API"""
+        self._test_template(
```
folmos-at-orange, line 2085: Document well the `source_type` in the docstring of `_test_template`.

folmos-at-orange (approved):
LGTM

One minor detail and you may merge.

tests/test_sklearn.py
popescu-v: Ok, done; updated the documentation.

popescu-v (approved):
LGTM.

PR #161, opened 2024-03-15
Use shell-based environment module handling for MPI detection

### LINKED ISSUE #157: Find a workaround for Rocky 8
#### Description
Currently, `khiops-python` doesn't work in Rocky 8. This is because the default python there is 3.6. We should find a workaround to support it.

#### Questions/Ideas
Options
- Make a fallback strategy for the `modules` part of the runner
- Make it work by calling bash
- Propose an installation for Khiops > `3.8`.

popescu-v:
closes #157


khiops/core/internals/runner.py
```diff
@@ -1051,88 +1048,71 @@ def _initialize_mpi_command_args(self):
...
             module_init_script_path = os.path.join(
                 os.path.sep, "etc", "profile.d", "modules.sh"
             )
-            modules_home_path = None
+
+            # List available environment modules
             if os.path.exists(module_init_script_path):
                 shell_command = shlex.split(
-                    f"bash -c 'source {module_init_script_path} && env'"
+                    f"bash -c 'source {module_init_script_path} && module avail'"
```
folmos-at-orange, line 1058: - Is "modules.sh" made for run with bash ?
  - If not I think it should be run with `sh`
- Should we check that bash exists ?

folmos-at-orange (commented):
LGTM after:

- Adding `rocky8` to test this code.

khiops/core/internals/runner.py
popescu-v: According to the documentation, both Sh and Bash are supported:
- https://modules.readthedocs.io/en/latest/#environment-modules: "All popular shells are supported, including bash, ksh, zsh, sh, csh, tcsh, fish, cmd, as well as some scripting languages such as tcl, perl, python, ruby, cmake and r."
- https://modules.readthedocs.io/en/latest/INSTALL.html#configuration: "These profile scripts will automatically adapt to the kind of sh or csh shell you are running."

Hence, IIUC the script works in both Sh and  Bash. However, using sh would alleviate the need to check that Bash exists, indeed. So, I would change to `sh` instead of `bash`.

popescu-v: Replacement with `sh` done and it seems to work. Hence, this can be closed IMHO.

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -33,8 +45,15 @@ RUN true \
                 KhiopsNativeInterface norm_jar khiops_jar \
   && cmake --install ./build/linux-gcc-release \
   && cd .. \
-  # Set python to python3 \
-  && alternatives --install /usr/bin/python python /usr/bin/python3 1 \
+  # Set python to python3.11 and pip to Pip 3.11 on Rocky 8 \
+  # Set python to python3 on Rocky 9 \
+  && if [ "$KHIOPSDEV_OS" = "rocky8" ]; then \
+     alternatives --set python /usr/bin/python3.11 \
+     && alternatives --install /usr/bin/pip pip /usr/bin/pip3.11 1 ; \
```
folmos-at-orange, line 52: I think this `&&` isn't necessary

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -33,8 +45,15 @@ RUN true \
...
+     else \
+         alternatives --install /usr/bin/python python /usr/bin/python3 1 \
+         && alternatives --install /usr/bin/pip pip /usr/bin/pip3 1 ; \
+     fi \
```
folmos-at-orange, line 56: Well indent this. The indent level of the `if` should be the same as the `else`.

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -15,7 +17,17 @@ RUN true \
   && dnf install --enablerepo=devel -y \
     git \
     pandoc \
-    python3-pip \
+  # Install Python 3.11 if on Rocky 8 \
```
folmos-at-orange, line 20: Can we install python 3.9 ?

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -7,6 +7,8 @@ LABEL description="Container for the development of khiops-python"
 
 # Install dev tools; build and install Khiops; set mpich as the default MPI
 ARG KHIOPS_REVISION
+# Reuse KHIOPSDEV_OS from previous stage
+ARG KHIOPSDEV_OS
 RUN useradd -rm -d /home/rocky -s /bin/bash -g root -u 1000 rocky
```
folmos-at-orange, line 12: It is necessary to add layer for this ?

folmos-at-orange: Also, the `RUN` paragraph should be separated frmo the `ARG`

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -15,7 +17,17 @@ RUN true \
...
+  && if [ "$KHIOPSDEV_OS" = "rocky8" ]; then \
+     dnf install -y \
+     python3.11 \
+     python3.11-pip \
+     python3.11-setuptools ; \
+     else \
+        dnf install -y \
+        python3-setuptools \
+        python3-pip ; \
+     fi \
```
folmos-at-orange, line 30: Indent so the `if` block is the same level as the `else`.

folmos-at-orange (changes requested):

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: First off, on rocky 8, `pip` as such does not exist; we have `pip3`, `pip-3` and `pip3.6` which all stand for the latter.
Second, on rocky 9, without this, we get the following failure: https://github.com/KhiopsML/khiops-python/actions/runs/8343320281/job/22833298119#step:4:31

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: Yes, ok.

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: `KHIOPS_DEV` being an `ARG` defined before a `FROM` it is erased after the `FROM`; to recover it, we need to redefine it (see https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact)

popescu-v: > Also, the `RUN` paragraph should be separated frmo the `ARG`

OK for adding an enter after the `ARG`.

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: Yes, ok.

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: Yes, we can. But Python 3.11 allowed us to uncover the issue with the version of `wrapt`.

packaging/docker/khiopspydev/Dockerfile.rocky
folmos-at-orange: Ok.

packaging/docker/khiopspydev/Dockerfile.rocky
folmos-at-orange: Ok

packaging/docker/khiopspydev/Dockerfile.rocky
folmos-at-orange: > `KHIOPS_DEV` being an `ARG` defined before a `FROM` it is erased after the `FROM`; to recover it, we need to redefine it (see https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact)

I meant the layer generated by `RUN useradd ...`

popescu-v: Oh, I see. Well, I wanted to do a minimal modification pertaining as much as possible to the topic of the PR (as long as the previous state was neither buggy, nor insufficient wrt. the PR).
But IMHO we could have all this in the same `RUN`, indeed.

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: Done.

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: Done.

folmos-at-orange (approved):

Issue #152, opened 2024-02-23
Run unittest for every supported python version.


folmos-at-orange:
#### Description
Currently, we only run the unittests for python 3.10. We should do it for every python version supported (3.8 -> 3.12).

#### Questions/Ideas
- Add miniconda to the dev docker with 5 envs (one for each python) and mpich installed.
- Remove from the dev docker mpich system-wide installation 
- Make the python version a matrix in the CI

popescu-v:
IMHO, we should also keep the out-of-Conda-environment tests (and the system-wide MPICH), in order to keep testing the `KhiopsLocalRunner` behaviour on genuine system-wide installations (especially wrt. MPI detection integration tests).

folmos-at-orange:
> IMHO, we should also keep the out-of-Conda-environment tests (and the system-wide MPICH), in order to keep testing the `KhiopsLocalRunner` behaviour on genuine system-wide installations (especially wrt. MPI detection integration tests).

Right, then we should have the following variants of `khiopspydev`:
- For systemwide the following OSs: `rocky8`, `rocky9`, `ubuntu22.04`, `ubuntu20.04`
- For conda: `conda-py38`, `conda-py39`, `conda-py310`, `conda-py311`, `conda-py312`

PR #162, opened 2024-03-18
Prepare multiple python3 envs (via miniconda3) in the `khiopspydev` docker image in order to run the unit tests on the supported versions

tramora:
- miniconda3 was chosen to keep the docker image light
- the conda envs follow this name pattern py$version (where version is the python3 version)
- in the github workflow the last two steps "Unit Tests" and "Upload Test Reports as Artifacts" may break because of a new wildcard in the folder name


tramora:
My fear was confirmed : the "dorny/test-reporter" plugin called in the github workflow does not handle a wildcard (*) in the folder name. I am working on this blocking issue introduced by this PR

tramora:
For some obscure reason on a container created by github action, the command 
```
Run python setup.py egg_info
```
finds the bogus string "python" as a version.
```
/usr/lib/python3/dist-packages/setuptools/dist.py:507: UserWarning: The version specified ('python') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
```

Under miniconda3 it is not just a warning but an exception that makes the command exit.

This strange behaviour does not show when **testing exactly the same docker image** locally !

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -24,6 +24,11 @@ RUN true \
   && cd .. \
   # Set python to python3 \
   && update-alternatives --install /usr/bin/python python /usr/bin/python3 1 \
+  # Install conda (miniconda3) to have multiple python3 versions \
+  && mkdir -p /root/miniconda3 \
+  && wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O /root/miniconda3/miniconda.sh \
+  && bash /root/miniconda3/miniconda.sh -b -u -p /root/miniconda3 \
+  && rm -rf /root/miniconda3/miniconda.sh \
```
popescu-v, line 31: Remove MPICH by adding (as we should be using Conda-based MPICH to be compatible with the Python version and the libc):
```sh
apt-get remove --purge libmpich-dev
```

popescu-v (changes requested):
Several changes are proposed (see comments).

tramora:
Thank you for your first feedback @popescu-v

.github/workflows/dev-docker.yml
```diff
@@ -60,6 +60,7 @@ jobs:
             "KHIOPS_REVISION=${{ env.KHIOPS_REVISION }}"
             "KHIOPSDEV_OS=${{ matrix.khiopsdev-os }}"
             "SERVER_REVISION=${{ env.SERVER_REVISION }}"
+            "PYTHON3_SUPPORTED_VERSIONS=3.8 3.9 3.10 3.11 3.12"
```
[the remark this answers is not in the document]
tramora: OK to enclose the versions list with quotes or double-quotes. I had a bad surprise during my tests to get an extra leading quote for the first element and an extra leading one for the last element.

I'll take into account the remaining remarks as well

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +55,25 @@ jobs:
...
         if: success() || failure()
         run: |
+          source /root/miniconda3/bin/activate py${{ matrix.python3-supported-versions }}  # warning : conda activate fails
           python setup.py egg_info
           pip install `grep -v "^\[" khiops.egg-info/requires.txt`
           rm -rf khiops.egg-info
-      - name: Prepare Unit Tests Environment
+        shell: bash -e {0}  # "bash" param value is translated internally into "bash --noprofile --norc -eo pipefail {0}"
+                            # that makes the step fail
+                            # because for an obscure reason "python setup.py egg_info" finds the boggus "python" str as a version
```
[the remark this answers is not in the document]
tramora: Yes, this will be removed, it was an (unsuccessful) test to figure out why the "python setup egg_info" would fail on the remote container.

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +55,25 @@ jobs:
...
           pip install unittest-xml-reporting
           pip install -r test-requirements.txt
-      - name: Install khiops-python dependencies
+        shell: bash
+      - name: Install khiops-python dependencies for each python3 version
         if: success() || failure()
         run: |
+          source /root/miniconda3/bin/activate py${{ matrix.python3-supported-versions }}  # warning : conda activate fails
           python setup.py egg_info
           pip install `grep -v "^\[" khiops.egg-info/requires.txt`
```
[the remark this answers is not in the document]
popescu-v: > Yes I was studying this part but also feared that removing system wide mpich would break the other integration tests `check-mpiexec-on-linux`

Yes, good point. Hence, given that, inside a Conda environment, the Conda-installed MPICH takes precedence over the system-wide MPICH installation, I think we can just install MPICH within Conda and also keep the system-wide installation. The khiops-python local runner should detect that we are in a Conda environment anyway, and use the Conda-installed MPICH.

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -55,11 +55,23 @@ RUN true \
        alternatives --install /usr/bin/python python /usr/bin/python3 1 \
        && alternatives --install /usr/bin/pip pip /usr/bin/pip3 1 ; \
      fi \
+  # Install conda (miniconda3) to have multiple python3 versions \
+  && mkdir -p /root/miniconda3 \
+  && wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O /root/miniconda3/miniconda.sh \
+  && bash /root/miniconda3/miniconda.sh -b -u -p /root/miniconda3 \
+  && rm -rf /root/miniconda3/miniconda.sh \
```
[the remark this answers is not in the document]
popescu-v: > I think we should remove system-wide MPICH (see issue #152 statement), by adding:
> 
> ```
> dnf remove mpich-devel
> ```

On second thought, following comment, we should keep the system-wide MPICH installation, but in this case, we should also keep the former system-wide tests, _beside_ the within-Conda environment tests.

packaging/docker/khiopspydev/Dockerfile.ubuntu
popescu-v: > I think we should remove system-wide MPICH by adding (as we should be using Conda-based MPICH to be compatible with the Python version and the libc):
> 
> ```shell
> apt-get remove --purge libmpich-dev
> ```

On second thought, IMHO we should keep the system-wide installation (plus the out-of-Conda env tests as well) - see previous comment in this respect.

popescu-v:
> For some obscure reason on a container created by github action, the command
> 
> ```
> Run python setup.py egg_info
> ```
> 
> finds the bogus string "python" as a version.
> 
> ```
> /usr/lib/python3/dist-packages/setuptools/dist.py:507: UserWarning: The version specified ('python') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
> ```
> 
> Under miniconda3 it is not just a warning but an exception that makes the command exit.
> 
> This strange behaviour does not show when **testing exactly the same docker image** locally !

This has now been fixed, by setting the GitHub repo as safe (as it is not situated in a directory that the current user owns).

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +56,26 @@ jobs:
...
+      - name: Setup and Install Test Requirements for each Python 3 version
         if: success() || failure()
         run: |
-          mkdir -p -m u+rwx ${{ github.workspace }}/reports
-          pip install unittest-xml-reporting
-          pip install -r test-requirements.txt
-      - name: Install khiops-python dependencies
+          mkdir -p -m u+rwx ${{ github.workspace }}/reports/py${{ matrix.python3-supported-versions }}            
+          # install within the conda environment without activating it 
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} pip install unittest-xml-reporting
```
popescu-v, line 64: Why does `conda install ... unittest-xml-reporting` not work?

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +56,26 @@ jobs:
...
+          /root/miniconda3/bin/conda install -y -n py${{ matrix.python3-supported-versions }} --file test-requirements.txt
+      - name: Install khiops-python dependencies for each Python 3 version
         if: success() || failure()
         run: |
-          python setup.py egg_info
-          pip install `grep -v "^\[" khiops.egg-info/requires.txt`
+          # required otherwise python setup.py egg_info would fail on github
```
popescu-v, line 69: Replace with something like:
```sh
# Required, as the Git repository is in a directory the current user does not own,
```

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +56,26 @@ jobs:
...
+          git config --global --add safe.directory $(realpath .)        
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} python setup.py egg_info
+          /root/miniconda3/bin/conda install -y -n py${{ matrix.python3-supported-versions }} `grep -v "^\[" khiops.egg-info/requires.txt`
           rm -rf khiops.egg-info
-      - name: Prepare Unit Tests Environment
+      - name: Install MPICH for each Python 3 version
+        if: success() || failure()
+        run: |
+          /root/miniconda3/bin/conda install -y -n py${{ matrix.python3-supported-versions }} mpich=4.1.1
```
popescu-v, line 77: Why MPICH 4.1.1. ? In the `khiops-core` Conda package, MPICH is pinned to 4.0.3 (see https://github.com/KhiopsML/khiops/pull/214/files#diff-0e83d52cc5466677512e7c46faff71d790b4765fbdc29afa5444cd2efc3ce6fe).

.github/workflows/unit-tests.yml
```diff
@@ -76,24 +87,24 @@ jobs:
...
+          git config --global --add safe.directory $(realpath .)                                
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} coverage run -m xmlrunner -o "reports/py${{ matrix.python3-supported-versions }}" -v
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} coverage report -m
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} coverage xml -o "reports/py${{ matrix.python3-supported-versions }}/py-coverage.xml"
+      - name: Display Unit Test Reports for each Python 3 version
         uses: dorny/test-reporter@v1
         with:
-          name: Unit Tests
-          path: reports/TEST-tests.*.*.xml
+          name: Unit Tests ${{ matrix.python3-supported-versions }}
```
popescu-v, line 97: I am not sure we need to change the stage name.

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -55,11 +55,25 @@ RUN true \
        alternatives --install /usr/bin/python python /usr/bin/python3 1 \
        && alternatives --install /usr/bin/pip pip /usr/bin/pip3 1 ; \
      fi \
+  # Install conda (miniconda3) to have multiple python3 versions \
+  && mkdir -p /root/miniconda3 && cd /root/miniconda3 \
+  && wget -qO- https://repo.anaconda.com/miniconda/ | pandoc -f html -t plain | awk '{print $5" "$1}' > ./sums \
+  && wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ./Miniconda3-latest-Linux-x86_64.sh \
+  && sha256sum --ignore-missing -c ./sums \
```
popescu-v, line 62: Why `--ignore-missing`?

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -55,11 +55,23 @@ RUN true \
...
   && dnf clean all \
   && rm -rf ./khiops \
   && true
 
+# set up all the supported python3 environments (+pip) under conda (for the unit tests)
+# relying on a variable containing all the versions
+ARG PYTHON3_SUPPORTED_VERSIONS
+RUN /root/miniconda3/bin/conda init \
+    && for version in ${PYTHON3_SUPPORTED_VERSIONS}; \
+    do /root/miniconda3/bin/conda create -y -n py${version} python=${version} pip; done
```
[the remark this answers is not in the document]
popescu-v: What happens if we do not use `pip install` anywhere? Also see my previous comment regarding `unittest-xml-reporting`.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -32,6 +39,13 @@ RUN true \
   && rm -rf ./khiops \
   && true
 
+# set up all the supported Python 3 environments (+pip) under conda (for the unit tests)
+# relying on a variable containing all the versions
+ARG PYTHON3_SUPPORTED_VERSIONS
+RUN /root/miniconda3/bin/conda init \
```
popescu-v, line 45: AFAIK, we don't need `conda init` anymore (see above comment in this respect as well).

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +55,23 @@ jobs:
           ref: ${{ env.SAMPLES_REVISION }}
           token: ${{ secrets.GITHUB_TOKEN }}
           path: khiops-samples
-      - name: Setup and Install Test Requirements
+      - name: Setup and Install Test Requirements for each python3 version
```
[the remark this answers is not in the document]
popescu-v: Put Python with capital "P".

.github/workflows/dev-docker.yml
```diff
@@ -3,6 +3,7 @@ name: Dev Docker
 env:
   DEFAULT_KHIOPS_REVISION: main
   DEFAULT_SERVER_REVISION: main
+  DEFAULT_PYTHON3_SUPPORTED_VERSIONS: 3.8 3.9 3.10 3.11 3.12
```
popescu-v, line 6: As previously stated, I would quote these versions.

.github/workflows/dev-docker.yml
```diff
@@ -20,6 +21,10 @@ on:
         type: boolean
         default: true
         description: Push to GH Registry
+      python3-supported-versions:
+        type: string
+        default: 3.8 3.9 3.10 3.11 3.12
```
popescu-v, line 26: Quote the Python versions list.

popescu-v (changes requested):
See remaining comments.

.github/workflows/dev-docker.yml
tramora: My bad. I forgot it

.github/workflows/dev-docker.yml
tramora: My bad. I forgot it

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -24,6 +24,13 @@ RUN true \
   && cd .. \
   # Set python to python3 \
   && update-alternatives --install /usr/bin/python python /usr/bin/python3 1 \
+  # Install conda (miniconda3) to have multiple Python 3 versions \
+  && mkdir -p /root/miniconda3 && cd /root/miniconda3 \
+  && wget -qO- https://repo.anaconda.com/miniconda/ | pandoc -f html -t plain | awk '{print $5" "$1}' > ./sums \
+  && wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ./Miniconda3-latest-Linux-x86_64.sh \
+  && sha256sum --ignore-missing -c ./sums \
```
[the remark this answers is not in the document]
tramora: Fetching from the https://repo.anaconda.com/miniconda/ URL gives all the releases with the corresponding checksums.

Without the `--ignore-missing` flag `sha256sum` will search locally for every file in the list and will fail for every missing one.

As we do use only one release (`Miniconda3-latest-Linux-x86_64.sh`), the flag is required to get a zero return code in `$?`

.github/workflows/unit-tests.yml
tramora: It does not find the package in the default channel. Maybe specifying `conda-forge` as you suggested will let me get rid of pip. I'll try it

.github/workflows/dev-docker.yml
tramora: In fact I did not forget but 'yamlfix' called by the pre-commit hook removes any quote or double quote (even if the yaml was valid) other linters seem to behave the same way (https://www.yamllint.com/)

.github/workflows/dev-docker.yml
tramora: In fact I did not forget but 'yamlfix' called by the pre-commit hook removes any quote or double quote (even if the yaml was valid) other linters seem to behave the same way (https://www.yamllint.com/)

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: OK, got this. Basically, we check that the current release's checksums is _among_ the checksums that are fetched from Anaconda (and that are re-fetched upon each image rebuild). But we do not check that the checksum is precisely as stated.
In order to ensure reproducible environments, IMHO we need to hard-code the checksum for the release that we want here. But for this, I would not use `latest`, but `Miniconda3-py312_24.1.2-0-Linux-x86_64.sh`, which currently corresponds to the `latest`. Hence, I would freeze both the miniconda version _and_ its checksum.

packaging/docker/khiopspydev/Dockerfile.ubuntu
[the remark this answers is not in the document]
popescu-v: Yes, but see my above comment in this respect.

popescu-v (approved):
LGTM

.github/workflows/dev-docker.yml
```diff
@@ -20,6 +21,10 @@ on:
         type: boolean
         default: true
         description: Push to GH Registry
+      python3-supported-versions:
```
folmos-at-orange, line 24: I'd rename to `python-versions`

.github/workflows/dev-docker.yml
```diff
@@ -20,6 +21,10 @@ on:
...
+        type: string
+        default: 3.8 3.9 3.10 3.11 3.12
+        description: Python versions supported by Khiops Python
```
folmos-at-orange, line 27: `Khiops Python` -> `khiops-python`

.github/workflows/unit-tests.yml
```diff
@@ -20,6 +20,10 @@ concurrency:
 jobs:
   run:
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python3-supported-versions: ['3.8', '3.9', '3.10', '3.11', '3.12']
```
folmos-at-orange, line 26: I'd rename to `python-version`

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +56,23 @@ jobs:
           ref: ${{ env.SAMPLES_REVISION }}
           token: ${{ secrets.GITHUB_TOKEN }}
           path: khiops-samples
-      - name: Setup and Install Test Requirements
+      - name: Setup and Install Test Requirements for each Python 3 version
```
folmos-at-orange, line 59: I'd remove the 'for each Python 3 version'. It is implicit that we are testing for a specific python version.

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +56,23 @@ jobs:
...
         if: success() || failure()
         run: |
-          mkdir -p -m u+rwx ${{ github.workspace }}/reports
-          pip install unittest-xml-reporting
-          pip install -r test-requirements.txt
-      - name: Install khiops-python dependencies
+          mkdir -p -m u+rwx ${{ github.workspace }}/reports/py${{ matrix.python3-supported-versions }}
```
folmos-at-orange, line 62: I don't think tha `${{ github.workspace }}`` is necessary here, just the relative path.

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +56,23 @@ jobs:
...
-          python setup.py egg_info
-          pip install `grep -v "^\[" khiops.egg-info/requires.txt`
+          # Required, as the Git repository is in a directory the current user does not own,
+          # Python versioneer fails to compute the current version correctly otherwise
+          git config --global --add safe.directory $(realpath .)        
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} python setup.py egg_info
+          /root/miniconda3/bin/conda install -y -n py${{ matrix.python3-supported-versions }} `grep -v "^\[" khiops.egg-info/requires.txt`
           rm -rf khiops.egg-info
-      - name: Prepare Unit Tests Environment
+      - name: Prepare Unit Tests Environment for each Python 3 version
```
folmos-at-orange, line 75: I'd remove the `for each...`. See above.

.github/workflows/unit-tests.yml
```diff
@@ -76,24 +84,24 @@ jobs:
         run: |
           # This is needed so that the Git tag is parsed and the khiops-python
           # version is retrieved
-          git config --global --add safe.directory $(realpath .)
-          coverage run -m xmlrunner -o "reports" -v
-          coverage report -m
-          coverage xml -o "reports/py-coverage.xml"
-      - name: Display Unit Test Reports
+          git config --global --add safe.directory $(realpath .)
```
folmos-at-orange, line 87: I'd remove the trailing whitespaces

.github/workflows/unit-tests.yml
```diff
@@ -52,19 +56,23 @@ jobs:
...
         run: |
-          mkdir -p -m u+rwx ${{ github.workspace }}/reports
-          pip install unittest-xml-reporting
-          pip install -r test-requirements.txt
-      - name: Install khiops-python dependencies
+          mkdir -p -m u+rwx ${{ github.workspace }}/reports/py${{ matrix.python3-supported-versions }}            
+          # install within the conda environment without activating it 
+          /root/miniconda3/bin/conda install -y -n py${{ matrix.python3-supported-versions }} -c conda-forge unittest-xml-reporting
+          /root/miniconda3/bin/conda install -y -n py${{ matrix.python3-supported-versions }} --file test-requirements.txt
+      - name: Install khiops-python dependencies for each Python 3 version
```
folmos-at-orange, line 66: I'd remove the `for each...`. See above.

.github/workflows/unit-tests.yml
```diff
@@ -76,24 +84,24 @@ jobs:
...
-          git config --global --add safe.directory $(realpath .)
-          coverage run -m xmlrunner -o "reports" -v
-          coverage report -m
-          coverage xml -o "reports/py-coverage.xml"
-      - name: Display Unit Test Reports
+          git config --global --add safe.directory $(realpath .)                                
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} coverage run -m xmlrunner -o "reports/py${{ matrix.python3-supported-versions }}" -v
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} coverage report -m
+          /root/miniconda3/bin/conda run --no-capture-output -n py${{ matrix.python3-supported-versions }} coverage xml -o "reports/py${{ matrix.python3-supported-versions }}/py-coverage.xml"
+      - name: Display Unit Test Reports for each Python 3 version
```
folmos-at-orange, line 91: I'd remove the `for each...`. See above.

.github/workflows/unit-tests.yml
```diff
@@ -76,24 +84,24 @@ jobs:
...
         uses: dorny/test-reporter@v1
         with:
-          name: Unit Tests
-          path: reports/TEST-tests.*.*.xml
+          name: Unit Tests ${{ matrix.python3-supported-versions }}
+          path: reports/py${{ matrix.python3-supported-versions }}/TEST-tests.*.*.xml
           reporter: java-junit
           path-replace-backslashes: 'true'  # Necessary for windows paths
-      - name: Upload Test Reports as Artifacts
+      - name: Upload Test Reports as Artifacts for each Python 3 version
```
folmos-at-orange, line 98: I'd remove the `for each...`. See above.

.github/workflows/unit-tests.yml
```diff
@@ -128,6 +136,9 @@ jobs:
           fetch-depth: 0
       - name: Install khiops-python dev dependencies
         run: |
+          # Required, as the Git repository is in a directory the current user does not own,
```
folmos-at-orange, line 139: The following git comman is required, as ...

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -8,8 +8,7 @@ LABEL description="Container for the development of khiops-python"
 # Reuse KHIOPSDEV_OS from previous stage
 ARG KHIOPSDEV_OS
 ARG KHIOPS_REVISION
-
-# Install dev tools; build and install Khiops; set mpich as the default MPI
+# Install dev tools and miniconda3 (for the unit tests); build and install Khiops; set mpich as the default MPI
```
folmos-at-orange, line 11: Use bullet points.

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -55,11 +55,25 @@ RUN true \
        alternatives --install /usr/bin/python python /usr/bin/python3 1 \
        && alternatives --install /usr/bin/pip pip /usr/bin/pip3 1 ; \
      fi \
+  # Install conda (miniconda3) to have multiple Python 3 versions \
```
folmos-at-orange, line 58: I'd choose between `conda` or `miniconda3`

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -55,11 +55,25 @@ RUN true \
...
+    && mkdir -p /root/miniconda3 && cd /root/miniconda3 \
+    # In case we would like to use the latest miniconda3 release, all the file names and their corresponding checksums \
+    # are listed on https://repo.anaconda.com/miniconda/ (list to be converted to plain text using pandoc) \
```
folmos-at-orange, line 61: I think the version will be pinned anyway so the "pandoc" comment is innecessary.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -32,6 +40,12 @@ RUN true \
   && rm -rf ./khiops \
   && true
 
+# set up all the supported Python 3 environments under conda (for the unit tests)
+# relying on a variable containing all the versions
+ARG PYTHON3_SUPPORTED_VERSIONS
```
folmos-at-orange, line 45: Same comments as the other dockerfile.

folmos-at-orange (changes requested):
Small editorial changes.

Also shorten the title of the PR **and** the commit. Something like "Install conda in dev Docker's for all supported python versions"

You may explain in the body of the message of the commit that it is to run the unit tests.

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: How about "Install miniconda3 to have multiple Python 3 versions via Conda"?

folmos-at-orange: "Install miniconda3 to have envs for each python version"

folmos-at-orange: Python 2 isn't relevant anymore so it is just Python.

.github/workflows/unit-tests.yml
popescu-v: Yeah, it's just that it was like that before, given as absolute path.
The semantic difference I see between using versus not using the `${{ github.workspace }}` prefix path is that by using it, the reports directory will always be relative to the _default_ working directory for GitHub jobs and stages; whereas, with a relative path, the reports directory location might change, depending on the per-job `[defaults: run:] working-directory` setting.

popescu-v: And in our case here, IMHO we want the location of the reports to be customizable by the user, via the `working-directory` setting. Hence, I would indeed use a relative path here.

.github/workflows/unit-tests.yml
folmos-at-orange: (without `s`)

.github/workflows/unit-tests.yml
folmos-at-orange: 1) The `working-directory` setting isn't used in the step (or job).
2) The step is executed always in the default working directory. Unless you need its full path it isn't necessary tu put `${{ github.workspace }}``.

.github/workflows/unit-tests.yml
tramora: Corrected (I get it the semantic is "the current version taken from the matrix")

.github/workflows/unit-tests.yml
popescu-v: Yes, whence my upvote to this proposed change.

folmos-at-orange (approved):

PR #170, opened 2024-03-26
Separate conda package into khiops-core and khiops

### LINKED ISSUE #168: Separate conda package in two
#### Description
Currently, the conda package contains the python code and the Khiops binaries. This makes building times very long as it recompiles khiops for each python version even though it is not necessary. Separating the binaries would speed up the tests and the release procedure.

#### Questions/Ideas
- Separate into the packages:
  - `khiops-core`: binaries 4 variants (Windows, Linux, Mac and Mac ARM)
  - `khiops`: python package depends on `khiops-core`
- This is more coherent with the current organization since a pure conda release would belong to the `khiops` CI.

folmos-at-orange:
Closes #168


.github/workflows/conda.yml
```diff
@@ -52,34 +37,19 @@ jobs:
...
-      # In Linux/macOS we need the conda-forge channel to install their pinned versions
-      - name: Build Khiops Conda Package (Linux/macOS)
+        run: conda install conda-build
+      - name: Build the Conda Package
         if: runner.os != 'Windows'
         run: |
           mkdir khiops-conda
-          conda build --channel conda-forge --output-folder khiops-conda ./packaging/conda
-      - name: Upload Khiops Conda Package
+          conda build --channel conda-forge --channel khiops-dev \
```
popescu-v, line 47: IMHO, the channel for the upstream `khiops-core` dependency should be customizable, between `khiops-dev` and `khiops`, so that one can build against an officially released `khiops-core` Conda package when building the "final" `khiops` Python Conda package.

.github/workflows/conda.yml
```diff
@@ -116,15 +86,16 @@ jobs:
...
       - name: Install the Khiops Conda pagkage (Windows)
         if: runner.os == 'Windows'
-        run: conda install -c ./khiops-conda/ khiops
+        run: conda install --channel khiops-dev --channel ./khiops-conda/ khiops
       # In Linux/macOS we need the conda-forge channel to install their pinned versions
       - name: Install the Khiops Conda package (Linux/macOS)
         if: runner.os != 'Windows'
-        run: conda install -c conda-forge -c ./khiops-conda/ khiops
+        run: |
+          conda install --channel conda-forge --channel khiops-dev --channel ./khiops-conda/ khiops
```
popescu-v, line 98: I would make the channel for the upstream `khiops-core` Conda package customizable, between `khiops-dev` and `khiops` (see my previous comment in this respect for the rationale).

.github/workflows/conda.yml
```diff
@@ -140,27 +111,26 @@ jobs:
...
-      - name: Install requirement packages
-        run: conda install -y anaconda-client conda-build=3.27.0
+          python-version: '3.12'
+      - name: Install Requirement Packages
+        run: conda install -y anaconda-client conda-index
       - name: Reindex the package directory
-        run: conda-index ./khiops-conda
-      - name: Upload the packages to anaconda.org
+        run: python -m conda_index ./khiops-conda
+      - name: Upload the Package to anaconda.org
```
popescu-v, line 133: s/Package/package/ (no need for capital "P" here)

.github/workflows/conda.yml
```diff
@@ -169,13 +139,13 @@ jobs:
...
-              --user "$ANACONDA_CHANNEL" ./khiops-conda/*/*.tar.bz2
+              --user "$ANACONDA_CHANNEL" ./khiops-conda/noarch/*.tar.bz2
           # For the dev channel: upload with forcing
           else
             anaconda --token "${{ secrets.KHIOPS_DEV_ANACONDA_CHANNEL_TOKEN }}" upload \
-              --user "$ANACONDA_CHANNEL" --force ./khiops-conda/*/*.tar.bz2
+              --user "$ANACONDA_CHANNEL" --force ./khiops-conda/noarch/*.tar.bz2
           fi
-      - name: Extract package version
+      - name: Extract the Package Version
```
popescu-v, line 148: No need to capitalize "Package" here IMHO

packaging/conda/meta.yaml
```diff
@@ -7,68 +7,26 @@ source:
...
   entry_points:
     - kh-status = khiops.tools:kh_status_entry_point
     - kh-samples = khiops.tools:kh_samples_entry_point
     - kh-download-datasets = khiops.tools:kh_download_datasets_entry_point
     - pk-status = khiops.tools:pk_status_entry_point   # deprecated
-  number: 0
-  # Binary relocation of MODL and MODL_Coclustering is done in build.sh script
-  # This is to be able to sign it, see the script for more details.
-  # Only done when "KHIOPS_APPLE_CERTIFICATE_BASE64" is defined in the environment.
-  {% if "KHIOPS_APPLE_CERTIFICATE_COMMON_NAME" in os.environ %}
```
popescu-v, line 31: How would this be handled now, when `KHIOPS_APPLE_CERTIFICATE_COMMON_NAME` is set at the `khiops-core` packaging level? Don't we need to make these Khiops Python entry points relocatable anymore on Mac OS?

popescu-v (changes requested):

.github/workflows/conda.yml
folmos-at-orange: I don't think so. However the version may be customizable.

.github/workflows/conda.yml
folmos-at-orange: nitpick

.github/workflows/conda.yml
folmos-at-orange: se above.

packaging/conda/meta.yaml
folmos-at-orange: They are relocated by default.

.github/workflows/conda.yml
folmos-at-orange: See above.

.github/workflows/conda.yml
popescu-v: Oh, wait: actually, for _building_ the `khiops` Python Conda package we don't need the `khiops-core` Conda package, do we? (it is needed for _testing_ the `khiops` Python Conda package though). Hence, why do we need to keep `--channel khiops-dev` here?

.github/workflows/conda.yml
popescu-v: Here, for _installing_ / _testing_ the (just) build `khiops` Python Conda package, IMHO it is suitable to be able to choose to install an officially-released `khiops-core` Conda dependency (suppose the `khiops-core` with the right tag has been already released "yesterday" and "today" we run the workflows for building / testing / releasing the `khiops` package).

Hence, in the _testing_ context, I would like to be able to choose whether to install the upstream `khiops-core` dependency from the dev channel, vs. the official release (to reproduce the actual usage context as much as possible).

.github/workflows/conda.yml
folmos-at-orange: Indeed, it is only to make the basics tests. And I think testing should be done only with assets of the test channel (for simplicity). The official channel contains only assets that have been tested.

.github/workflows/conda.yml
folmos-at-orange: Any official package must have been passed by test. The `khiops-dev` channel will contain all versions `khiops-core` released. So we only need the version.

popescu-v: OK, if we are sure that the same-version `khiops-core` package on `khiops-dev` is identical "at-the-bit-level" to the same package on `khiops`.

.github/workflows/conda.yml
popescu-v: So, for this _build_ stage, could we remove it?

folmos-at-orange: I think it is still necessary for the `test` part of the `meta.yaml`

popescu-v: I am not sure, as we only test module imports, not execution. Could you try without it?

packaging/conda/meta.yaml
```diff
@@ -7,68 +7,26 @@ source:
...
-    {% if "KHIOPS_APPLE_CERTIFICATE_COMMON_NAME" in os.environ %}
-    - KHIOPS_APPLE_CERTIFICATE_COMMON_NAME # [osx]
-    # Only available when "KHIOPS_APPLE_CERTIFICATE_BASE64" is defined in the environment.
-    {% if "KHIOPS_APPLE_CERTIFICATE_BASE64" in os.environ %}
-    - KHIOPS_APPLE_CERTIFICATE_BASE64      # [osx]
-    - KHIOPS_APPLE_CERTIFICATE_PASSWORD    # [osx]
-    - KHIOPS_APPLE_TMP_KEYCHAIN_PASSWORD   # [osx]
-    {% endif %}
-    {% endif %}
+  number: 0
```
popescu-v, line 10: This will always be `0` for us, or should we set-up a protocol for updating it?

.github/workflows/conda.yml
```diff
@@ -1,14 +1,14 @@
...
   DEFAULT_SAMPLES_REVISION: main
+  DEFAULT_KHIOPS_CORE_VERSION: 10.2.1
 on:
   workflow_dispatch:
     inputs:
-      khiops-revision:
-        default: main
-        description: khiops repo revision
+      khiops-core-version:
+        default: 10.2.1
```
popescu-v, line 10: Should we have a hard-coded default here? Because IMHO one risks not bumping it as needed.

Perhaps setting this default as computed from the current Git tag? (like `git describe --tags | awk 'BEGIN {FS = "."; OFS = FS} {print $1, $2, $3}'`?

.github/workflows/conda.yml
folmos-at-orange: kh-status would fail if it doesn't find the binaries.

.github/workflows/conda.yml
folmos-at-orange: Too much complexity. I think is better to maintain this by hand as it doesn't change that often. I'll add it to the release checklist.

popescu-v: OK, fair enough.

.github/workflows/conda.yml
popescu-v: OK, I see. I wonder if we need to test the _Python_ package _build_ with `kh-status`. Perhaps we could just keep the imports as tests, as this would stay within the scope of the pure Python package features. And we would keep the `kh-status` for the testing workflow job?

folmos-at-orange: I prefer to leave that minimal check for building out of the workflow.

.github/workflows/conda.yml
folmos-at-orange: Its a bit too much to 'bit-by-bit'. At least it would be constructed from the same tag. Or even transferred directly from `khiops-dev`

.github/workflows/conda.yml
popescu-v: We already launch `kh-status` in the package `test` workflow job. Why do we need to redundantly keep it the after-build tests as well? Removing it would allow us to focus on the pure Python aspects: just test that the namespacing is set-up right.

.github/workflows/conda.yml
popescu-v: Yes, perhaps this is the way to go: add workflow stage for: 1/ fetching the package from `khiops-dev`; 2/ uploading it to |`khiops`?

folmos-at-orange: Too complicated. Just build from the same tag.

.github/workflows/conda.yml
folmos-at-orange: You want at least a minimal test when you build in local. A bit of redundancy is not harmful.

.github/workflows/conda.yml
popescu-v: Building from the same tag does not guarantee that the packages are identical for the same tag IMHO (tags can be moved, development containers might change). Copying the package from `khiops-dev` to `khiops` does provide this guarantee.

folmos-at-orange: That is true. I don't think making things more complex for this minor risk is worth it.

.github/workflows/conda.yml
folmos-at-orange: Ok

.github/workflows/conda.yml
popescu-v: Well, if through convention we could enforce that:

- no non-alpha tag should be overwritten, _ever_, in the  `khiops-core` repo;
- no alpha tag should _ever_ be used for getting a `khiops-core` package as a dependency of the `khiops` package.

then the current behavior seems acceptable to me, as a trade-off. But this seems acceptable only if the aforementioned two points are clearly stated (as cautionary notes) in the contributing guides of the respective repositories:

- the first point should be mentioned somewhere in the `khiops` repo, perhaps in https://github.com/KhiopsML/khiops/wiki/Release-Process?
- the second point should be mentioned in the `CONTRIBUTING.md` file of the current repo.

folmos-at-orange: I wrote a comment in the wiki for the first point.
The second point I put a note in the workflow.

.github/workflows/conda.yml
folmos-at-orange: I think this is obsolete with the pinned version of `khiops-core`

CONTRIBUTING.md
```diff
@@ -244,22 +244,43 @@ carefree while still trying to not add too many dependencies.
...
+  - Update `CHANGELOG.md`
+  - Update the default `khiops-core` version in [.github/workflows/conda.yml]
+- Git manipulations
+  - Merge the `dev` branch into `main`
+  - Tag the merge commit with the release version (see Versioning above)
+  - Rebase the `dev` branch onto `main`
+    - `git switch main`
+    - `git pull`
+    - `git switch dev`
+    - `git pull`
```
popescu-v, line 271: Don't we need a `git rebase main` here, while on `dev`?

CONTRIBUTING.md
```diff
@@ -244,22 +244,43 @@ carefree while still trying to not add too many dependencies.
...
+- Release issue and its related PR
+  - Update the API Docs if necessary
+  - Update `CHANGELOG.md`
+  - Update the default `khiops-core` version in [.github/workflows/conda.yml]
+- Git manipulations
+  - Merge the `dev` branch into `main`
+  - Tag the merge commit with the release version (see Versioning above)
+  - Rebase the `dev` branch onto `main`
+    - `git switch main`
+    - `git pull`
```
popescu-v, line 269: `git pull --tags` to also get the tag locally?

CONTRIBUTING.md
```diff
@@ -244,22 +244,43 @@ carefree while still trying to not add too many dependencies.
...
+    - `git switch dev`
+    - `git pull`
+    - `git push dev` (you need to remove the protections of `dev` for this step)
```
popescu-v, line 272: If `dev`'s `HEAD` changed following the rebase, don't we need to `git push -f` here?

CONTRIBUTING.md
```diff
@@ -244,22 +244,43 @@ carefree while still trying to not add too many dependencies.
...
+    - This is necessary to include the merge commit into master to calculate intermediary versions
+      with versioneer.
+- Workflows
+  - Execute the `Conda Package` workflow specifying:
+    - The release tag
```
popescu-v, line 277: What about `khiops-core-version`? Couldn't this be specified as well, to make sure it matches the release tag?

CONTRIBUTING.md
```diff
@@ -244,22 +244,43 @@ carefree while still trying to not add too many dependencies.
...
+    - `khiops` as the release channel
+  - Execute the `API Docs` worflow specifying "Deploy GH Pages".
```
popescu-v, line 279: s/worflow/workflow/

CONTRIBUTING.md
folmos-at-orange: Ok

CONTRIBUTING.md
folmos-at-orange: It is specified above.

CONTRIBUTING.md
folmos-at-orange: Ok

CONTRIBUTING.md
folmos-at-orange: I put a `git reset --hard`

CONTRIBUTING.md
folmos-at-orange: Done

CONTRIBUTING.md
folmos-at-orange: Done

CONTRIBUTING.md
folmos-at-orange: Done

CONTRIBUTING.md
```diff
@@ -244,22 +244,53 @@ carefree while still trying to not add too many dependencies.
...
+    - `git merge dev`
+  - Tag the merge commit with the release version (see Versioning above)
+    - `git switch main`
+    - `git tag 10.3.0.1 # Just an example`
+  - Make `dev` point to the merge commit just created in `main`
+    - This is necessary to include the merge commit into master to calculate intermediary versions
+      with Versioneer.
+    - Steps:
+      - `git switch dev`
+      - `git reset --hard main`
```
popescu-v, line 283: We avoid rebasing onto `main` so that we do not need to `git push -f dev`?

tests/test_core.py
```diff
@@ -404,20 +404,30 @@ def test_api_scenario_generation(self):
...
-                        else None,
-                        "output_additional_data_tables": {
-                            bytes(key, encoding="ascii"): bytes(value, encoding="ascii")
-                            for key, value in output_additional_data_tables[
-                                dataset
-                            ].items()
-                        }
-                        if output_additional_data_tables[dataset] is not None
-                        else None,
+                        "additional_data_tables": (
```
popescu-v, line 407: These Python code reformattings were all caused by the `black` version updating, right?

folmos-at-orange: Yes

CONTRIBUTING.md
folmos-at-orange: No, this command will just make `dev` point to `main`. The push is still necessary.

folmos-at-orange: And `--force` isn't necessary as we are adding a commit to the tip of the branch.

popescu-v (approved):
LGTM.

PR #199, opened 2024-07-02
Fix Conda package version extraction from Git tag

popescu-v:
The '-' pre-release version character is removed from the Conda version.

### DIFF
--- a/packaging/conda/meta.yaml
+++ b/packaging/conda/meta.yaml
@@ -1,7 +1,8 @@
 {% set metadata = load_setup_py_data(setup_file='../../setup.py', from_recipe_dir=True) %}
 package:
   name: {{ metadata.get('name') }}
-  version: {{ metadata.get('version') }}
+  # The Conda version cannot contain the '-' character, so we eliminate it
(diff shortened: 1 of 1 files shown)


packaging/conda/meta.yaml
```diff
@@ -1,7 +1,7 @@
 {% set metadata = load_setup_py_data(setup_file='../../setup.py', from_recipe_dir=True) %}
 package:
   name: {{ metadata.get('name') }}
-  version: {{ metadata.get('version') }}
+  version: {{ metadata.get('version') | replace('-', '') }}
```
folmos-at-orange, line 4: Just comment why we do this.

folmos-at-orange (changes requested):

packaging/conda/meta.yaml
popescu-v: Ok.

popescu-v: Done.

folmos-at-orange (approved):

Issue #175, opened 2024-04-09
Reactivate the S3 and GCS filesystem integration tests


folmos-at-orange:
#### Description
Currently, the S3 and GCS integration tests are not active since we have not configured the credentials needed. We should reactivate them and add more tests.

#### Questions/Ideas
- We have already a GCS bucket for that it. We'll just need the appropriate secrets.
- For S3 we may have need have a bucket accessible from GitHub.
- We'll need to update the `boto3` code and min version as the API changed.
- We need to test all the API operations.

popescu-v:
Tests done on fake GCS and S3 servers deployed "on-prems" in the GitHub CI.

Tests done for natively-installed `khiops-core` binary package and drivers. For Conda, drivers are not available yet.

popescu-v:
Closed by PR https://github.com/KhiopsML/khiops-python/pull/210.

Issue #53, opened 2023-07-04
GCS tests do not completely clean themselves


popescu-v:
When setting an `output_dir` in the `khiops_classifier` remote file tests the directory is left empty.

popescu-v:
Can it be addressed in PR #175 ?

popescu-v:
Closed by PR https://github.com/KhiopsML/khiops-python/pull/210.

PR #210, opened 2024-07-11
Reactivate the integration tests for remote file access (GCS, S3)

tramora:
Docker is installed in the khiopspydev docker image so that it can start the local fake servers .
The container must be started with the --privileged option otherwise docker inside the container would not start.
The fake servers are pre-provisioned with sample files to be ready before the tests run

closes https://github.com/KhiopsML/khiops-python/issues/175, https://github.com/KhiopsML/khiops-python/issues/53

### DIFF
--- a/.github/workflows/dev-docker.yml
+++ b/.github/workflows/dev-docker.yml
@@ -5,8 +5,8 @@ env:
   DEFAULT_IMAGE_INCREMENT: 0
   DEFAULT_SERVER_REVISION: main
   DEFAULT_PYTHON_VERSIONS: 3.8 3.9 3.10 3.11 3.12 3.13
-  DEFAULT_KHIOPS_GCS_DRIVER_REVISION: 0.0.10
-  DEFAULT_KHIOPS_S3_DRIVER_REVISION: 0.0.12
+  DEFAULT_KHIOPS_GCS_DRIVER_REVISION: 0.0.11
+  DEFAULT_KHIOPS_S3_DRIVER_REVISION: 0.0.13
 on:
   pull_request:
     paths: [packaging/docker/khiopspydev/Dockerfile.*, .github/workflows/dev-docker.yml]
@@ -38,11 +38,11 @@ on:
         description: Khiops Server Revision
       khiops-gcs-driver-revision:
         type: string
-        default: 0.0.10
+        default: 0.0.11
         description: Driver version for Google Cloud Storage remote files
       khiops-s3-driver-revision:
         type: string
-        default: 0.0.12
+        default: 0.0.13
         description: Driver version for AWS-S3 remote files
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/dev-docker.yml
+++ b/.github/workflows/dev-docker.yml
@@ -5,6 +5,8 @@ env:
   DEFAULT_IMAGE_INCREMENT: 0
   DEFAULT_SERVER_REVISION: main
   DEFAULT_PYTHON_VERSIONS: 3.8 3.9 3.10 3.11 3.12 3.13
+  DEFAULT_KHIOPS_GCS_DRIVER_REVISION: 0.0.10
+  DEFAULT_KHIOPS_S3_DRIVER_REVISION: 0.0.12
 on:
   pull_request:
     paths: [packaging/docker/khiopspydev/Dockerfile.*, .github/workflows/dev-docker.yml]
@@ -34,6 +36,14 @@ on:
         type: string
         default: main
         description: Khiops Server Revision
+      khiops-gcs-driver-revision:
+        type: string
+        default: 0.0.10
+        description: Driver version for Google Cloud Storage remote files
+      khiops-s3-driver-revision:
+        type: string
+        default: 0.0.12
+        description: Driver version for AWS-S3 remote files
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
@@ -55,6 +65,8 @@ jobs:
           echo "KHIOPSDEV_OS_CODENAME=$(echo '${{ matrix.khiopsdev-os }}' | tr -d '0-9.')" >> "$GITHUB_ENV"
           echo "SERVER_REVISION=${{ inputs.server-revision || env.DEFAULT_SERVER_REVISION }}" >> "$GITHUB_ENV"
           echo "IMAGE_URL=ghcr.io/khiopsml/khiops-python/khiopspydev-${{ matrix.khiopsdev-os }}" >> "$GITHUB_ENV"
+          echo "KHIOPS_GCS_DRIVER_REVISION=${{ inputs.khiops-gcs-driver-revision || env.DEFAULT_KHIOPS_GCS_DRIVER_REVISION }}" >> "$GITHUB_ENV"
+          echo "KHIOPS_S3_DRIVER_REVISION=${{ inputs.khiops-s3-driver-revision || env.DEFAULT_KHIOPS_S3_DRIVER_REVISION }}" >> "$GITHUB_ENV"
       - name: Checkout khiops-python sources
         uses: actions/checkout@v4
       - name: Set up Docker Buildx
@@ -81,6 +93,9 @@ jobs:
       - name: Build image and push it to GitHub Container Registry
         uses: docker/build-push-action@v5
         with:
+          # Special hostname used by the integration tests for remote file access
+          # added using inputs because /etc/hosts is read-only for alternate builders (buildx via moby buildkit)
+          add-hosts: s3-bucket.localhost:127.0.0.1
           context: ./packaging/docker/khiopspydev/
           file: ./packaging/docker/khiopspydev/Dockerfile.${{ env.KHIOPSDEV_OS_CODENAME }}
           build-args: |
@@ -88,6 +103,8 @@ jobs:
             "KHIOPSDEV_OS=${{ matrix.khiopsdev-os }}"
             "SERVER_REVISION=${{ env.SERVER_REVISION }}"
             "PYTHON_VERSIONS=${{ inputs.python-versions || env.DEFAULT_PYTHON_VERSIONS }}"
+            "KHIOPS_GCS_DRIVER_REVISION=${{ env.KHIOPS_GCS_DRIVER_REVISION }}"
+            "KHIOPS_S3_DRIVER_REVISION=${{ env.KHIOPS_S3_DRIVER_REVISION }}"
           tags: ${{ env.DOCKER_IMAGE_TAGS }}
           # Push only on manual request
           push: ${{ inputs.push || false }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,11 +15,11 @@ on:
       khiops-desktop-revision:
         default: 10.2.4
         description: Khiops Windows Desktop Application Version
-      run-long-tests:
+      run-expensive-tests:
         type: boolean
         required: false
         default: false
-        description: Execute long tests
+        description: Execute expensive tests
   pull_request:
     paths:
       - khiops/**.py
@@ -110,8 +110,8 @@ jobs:
             rm -rf khiops.egg-info
           done
       - name: Prepare Unit Tests Environment
-        if: github.ref != 'dev' && github.ref != 'main' && ! inputs.run-long-tests
-        run: echo "UNITTEST_ONLY_SHORT_TESTS=true" >> "$GITHUB_ENV"
+        if: github.ref != 'dev' && github.ref != 'main' && ! inputs.run-expensive-tests
+        run: echo "SKIP_EXPENSIVE_TESTS=true" >> "$GITHUB_ENV"
       - name: Prepare Integration Tests on remote files
         env:
           AWS_ENDPOINT_URL: http://localhost:4569
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -133,10 +133,19 @@ jobs:
           cat  ${GITHUB_WORKSPACE}/.aws/credentials
           echo "Generated AWS configuration..."
           cat  ${GITHUB_WORKSPACE}/.aws/configuration
-
-          # Prepare GCS credentials
-          touch ${GITHUB_WORKSPACE}/google-credentials.json
           /scripts/run_fake_remote_file_servers.sh .  # launch the servers in the background
+      - name: Authenticate to GCP using "Workload Identity Federation"
+        # For integration tests on GCS we use a real Google account
+        # Retrieve the Google credentials through "Workload Identity Federation"
+        # see https://github.com/google-github-actions/auth?tab=readme-ov-file#workload-identity-federation-through-a-service-account
+        uses: google-github-actions/auth@v2
+        with:
+          service_account: khiops-gcs-driver-test-sa@ino-olr-dak-ideal-sbx.iam.gserviceaccount.com
+          workload_identity_provider: projects/322269704080/locations/global/workloadIdentityPools/github/providers/my-repo
+          # 'create_credentials_file' is true by default but let's make it explicit
+          # After authentication, the required GOOGLE_APPLICATION_CREDENTIALS environment variable is exported
+          # https://github.com/google-github-actions/auth?tab=readme-ov-file#inputs-miscellaneous
+          create_credentials_file: true
       - name: Run Unit & Integration Tests
         env:
           KHIOPS_SAMPLES_DIR: ${{ github.workspace }}/khiops-samples
@@ -150,19 +159,14 @@ jobs:
           # Oversubscribe for MPI > 4.x
           OMPI_MCA_rmaps_base_oversubscribe: true
           # for the tests with GCS
-          GCS_BUCKET_NAME: gcs-bucket
-          # we take advantage of the built-in `STORAGE_EMULATOR_HOST` env variable
-          # that every GCS client can read and lets us use a local fake file server
-          STORAGE_EMULATOR_HOST: http://localhost:4443
-          # even in GCS emulation mode, the credentials file must exist
-          GOOGLE_APPLICATION_CREDENTIALS: ${{ github.workspace }}/google-credentials.json
+          GCS_BUCKET_NAME: data-test-khiops-driver-gcs/khiops_data
           GCS_DRIVER_LOGLEVEL: info  # set to debug for diagnosis
-          S3_DRIVER_LOGLEVEL: info  # set to debug for diagnosis
           # for the tests with S3
+          S3_DRIVER_LOGLEVEL: info  # set to debug for diagnosis
           S3_BUCKET_NAME: s3-bucket
           AWS_SHARED_CREDENTIALS_FILE: ${{ github.workspace }}/.aws/credentials
           AWS_CONFIG_FILE: ${{ github.workspace }}/.aws/configuration
-          # common var for tests with GCS & S3
+          # Var for tests with S3
           no_proxy: localhost
         run: |
           # This is needed so that the Git tag is parsed and the khiops-python
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,5 +1,5 @@
 ---
-name: Unit Tests
+name: Tests
 env:
   DEFAULT_SAMPLES_REVISION: 10.2.4
   DEFAULT_KHIOPS_DESKTOP_REVISION: 10.2.4
@@ -112,7 +112,32 @@ jobs:
       - name: Prepare Unit Tests Environment
         if: github.ref != 'dev' && github.ref != 'main' && ! inputs.run-long-tests
         run: echo "UNITTEST_ONLY_SHORT_TESTS=true" >> "$GITHUB_ENV"
-      - name: Run Unit Tests
+      - name: Prepare Integration Tests on remote files
+        env:
+          AWS_ENDPOINT_URL: http://localhost:4569
+        shell: bash
+        run: |
+          # Prepare AWS-S3 credentials and configuration          
+          mkdir -p ${GITHUB_WORKSPACE}/.aws/
+          cat << EOF > ${GITHUB_WORKSPACE}/.aws/credentials
+          [default]
+          aws_access_key_id=KEY
+          aws_secret_access_key=SECRET           
+          EOF
+          cat << EOF > ${GITHUB_WORKSPACE}/.aws/configuration
+          [default]
+          endpoint_url=${AWS_ENDPOINT_URL}
+          region=eu-north-1          
+          EOF
+          echo "Generated AWS credentials..."
+          cat  ${GITHUB_WORKSPACE}/.aws/credentials
+          echo "Generated AWS configuration..."
+          cat  ${GITHUB_WORKSPACE}/.aws/configuration
+
+          # Prepare GCS credentials
+          touch ${GITHUB_WORKSPACE}/google-credentials.json
+          /scripts/run_fake_remote_file_servers.sh .  # launch the servers in the background
+      - name: Run Unit & Integration Tests
         env:
           KHIOPS_SAMPLES_DIR: ${{ github.workspace }}/khiops-samples
           KHIOPS_DOCKER_RUNNER_URL: https://localhost:11000
@@ -124,6 +149,21 @@ jobs:
           rmaps_base_oversubscribe: true
           # Oversubscribe for MPI > 4.x
           OMPI_MCA_rmaps_base_oversubscribe: true
+          # for the tests with GCS
+          GCS_BUCKET_NAME: gcs-bucket
+          # we take advantage of the built-in `STORAGE_EMULATOR_HOST` env variable
+          # that every GCS client can read and lets us use a local fake file server
+          STORAGE_EMULATOR_HOST: http://localhost:4443
+          # even in GCS emulation mode, the credentials file must exist
+          GOOGLE_APPLICATION_CREDENTIALS: ${{ github.workspace }}/google-credentials.json
+          GCS_DRIVER_LOGLEVEL: info  # set to debug for diagnosis
+          S3_DRIVER_LOGLEVEL: info  # set to debug for diagnosis
+          # for the tests with S3
+          S3_BUCKET_NAME: s3-bucket
+          AWS_SHARED_CREDENTIALS_FILE: ${{ github.workspace }}/.aws/credentials
+          AWS_CONFIG_FILE: ${{ github.workspace }}/.aws/configuration
+          # common var for tests with GCS & S3
+          no_proxy: localhost
         run: |
           # This is needed so that the Git tag is parsed and the khiops-python
           # version is retrieved
@@ -138,10 +178,10 @@ jobs:
             $CONDA run --no-capture-output -n "$CONDA_ENV" coverage report -m
             $CONDA run --no-capture-output -n "$CONDA_ENV" coverage xml -o "reports/$CONDA_ENV/py-coverage.xml"
           done
-      - name: Display Unit Test Reports
+      - name: Display Test Reports
         uses: dorny/test-reporter@v1
         with:
-          name: Unit Tests ${{ matrix.python-version }}
+          name: Run Tests ${{ matrix.python-version }}
           path: >-
             reports/py${{ matrix.python-version }}/TEST-tests.*.*.xml,
             reports/py${{ matrix.python-version }}_conda/TEST-tests.*.*.xml
--- a/khiops/core/internals/filesystems.py
+++ b/khiops/core/internals/filesystems.py
@@ -23,6 +23,7 @@ from urllib.parse import urlparse
 try:
     import boto3
     import boto3.session
+    from boto3.exceptions import S3UploadFailedError
     from botocore.exceptions import ClientError
 
     boto3_import_error = None
@@ -254,7 +255,7 @@ def copy_from_local(uri_or_path, local_path):
     Raises
     ------
     RuntimeError
-        If there was a problem when removing.
+        If there was a problem when copying.
     """
     create_resource(uri_or_path).copy_from_local(local_path)
 
@@ -272,7 +273,7 @@ def copy_to_local(uri_or_path, local_path):
     Raises
     ------
     RuntimeError
-        If there was a problem when removing.
+        If there was a problem when copying.
     """
     create_resource(uri_or_path).copy_to_local(local_path)
 
@@ -668,29 +669,22 @@ class AmazonS3Resource(FilesystemResource):
             )
 
     def copy_from_local(self, local_path):
-        response = self.s3_client.Bucket(self.uri_info.netloc).upload_file(
-            local_path, self.uri_info.path[1:]
-        )
-        status_code = response["ResponseMetadata"]["HTTPStatusCode"]
-        copy_ok = 200 <= status_code <= 299
-        if not copy_ok:
-            raise RuntimeError(
-                f"S3 copy_from_local failed {self.uri} with code {status_code}: "
-                + json.dumps(response)
+        try:
+            self.s3_client.Bucket(self.uri_info.netloc).upload_file(
+                local_path, self.uri_info.path[1:]
             )
+        # normalize the raised exception
+        except S3UploadFailedError as exc:
+            raise RuntimeError(f"S3 copy_from_local failed {self.uri}") from exc
 
     def copy_to_local(self, local_path):
-        response = self.s3_client.Bucket(self.uri_info.netloc).download_file(
-            local_path, self.uri_info.path[1:]
-        )
-        status_code = response["ResponseMetadata"]["HTTPStatusCode"]
-        copy_ok = 200 <= status_code <= 299
-        if not copy_ok:
-            raise RuntimeError(
-                f"S3 download failed {self.uri} with code {status_code}: "
-                + json.dumps(response)
+        try:
+            self.s3_client.Bucket(self.uri_info.netloc).download_file(
+                self.uri_info.path[1:], local_path
             )
-        return copy_ok
+        # normalize the raised exception
+        except S3UploadFailedError as exc:
+            raise RuntimeError(f"S3 download failed {self.uri}") from exc
 
     def list_dir(self):
         # Add an extra slash to the path to treat it as a folder
--- a/packaging/conda/meta.yaml
+++ b/packaging/conda/meta.yaml
@@ -30,8 +30,12 @@ requirements:
     - pandas >=0.25.3
     - scikit-learn >=0.22.2
   run_constrained:
-    - boto3 >=1.17.39
+    # do not necessary use the latest version
+    # to avoid undesired breaking changes
+    - boto3 >=1.17.39,<=1.35.69
     - google-cloud-storage >=1.37.0
+    # an open issue on boto3 (https://github.com/boto/boto3/issues/3585) forces a min version of pyopenssl
+    - pyopenssl>=24.0.0,<25.0.0
 
 outputs:
   - name: {{ metadata.get('name') }}
--- a/packaging/docker/khiopspydev/Dockerfile.ubuntu
+++ b/packaging/docker/khiopspydev/Dockerfile.ubuntu
@@ -10,14 +10,15 @@ ARG KHIOPS_REVISION
 RUN true \
   # Install git (for khiops-python version calculation) and pip \
   && apt-get -y update \
-  && apt-get -y --no-install-recommends install git python3-pip zip pandoc wget \
+  && apt-get -y --no-install-recommends install git python3-pip zip pandoc wget ruby-dev \
   # Get Linux distribution codename \
   && if [ -f /etc/os-release ]; then . /etc/os-release; fi \
   # Obtain the Khiops native package \
   && KHIOPS_PKG_FILE=$KHIOPS_REVISION/khiops-core-openmpi_$KHIOPS_REVISION-1-$VERSION_CODENAME.amd64.deb \
   && wget -O KHIOPS_CORE.deb "https://github.com/KhiopsML/khiops/releases/download/${KHIOPS_PKG_FILE}" \
-  # Install the Khiops native package \
-  && dpkg -i --force-all KHIOPS_CORE.deb \
+  # Install the Khiops native package : make it always succeed. \
+  # If dpkg fails it is due to missing dependencies which will be installed by apt in the next line \
+  && (dpkg -i --force-all KHIOPS_CORE.deb || true) \
   && apt-get -f -y install \
   && rm -f KHIOPS_CORE.deb \
   # Set python to python3 \
@@ -53,10 +54,36 @@ RUN true \
   && true
 
 RUN mkdir -p /scripts
-COPY ./run_service.sh /scripts/run_service.sh
-RUN chmod +x /scripts/run_service.sh && \
+COPY ./run_service.sh ./run_fake_remote_file_servers.sh /scripts/
+RUN chmod +x /scripts/run_service.sh /scripts/run_fake_remote_file_servers.sh && \
   useradd -rm -d /home/ubuntu -s /bin/bash -g root -u 1000 ubuntu
 
+# new gcs & s3 drivers (written in C++)
+ARG KHIOPS_GCS_DRIVER_REVISION
+ARG KHIOPS_S3_DRIVER_REVISION
+RUN true \
+    && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_REVISION}/khiops-gcs_${KHIOPS_GCS_DRIVER_REVISION}.deb \
+    && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_REVISION}/khiops-s3_${KHIOPS_S3_DRIVER_REVISION}.deb \
+    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb || true) \
+    && apt-get -f -y install \
+    && rm -f khiops-gcs.deb khiops-s3.deb \
+    && true
+
 FROM ghcr.io/khiopsml/khiops-server:${SERVER_REVISION} AS server
+
 FROM khiopsdev AS base
 COPY --from=server /service /usr/bin/service
+
+# S3 fake file server (only in the ubuntu container)
+# Do not use the latest fakes3 version because starting from 1.3 a licence is required
+# if fakes3 is no longer compatible think about switching to an alternative and fully compatible server
+# (https://github.com/jamhall/s3rver:v3.7.1 is not yet for example)
+RUN gem install fakes3:1.2.1 sorted_set
+# Avoid resolving a fake s3-bucket.localhost hostname
+# Alternate builders (buildx via moby buildkit) mount /etc/hosts read-only, the following command will fail
+# echo "127.0.0.1 s3-bucket.localhost" >> /etc/hosts
+# You will have to add the `add-hosts` input instead (https://github.com/docker/build-push-action/#inputs)
+
+# Port on which fakes3 is listening
+EXPOSE 4569
+
--- a/packaging/docker/khiopspydev/Dockerfile.ubuntu
+++ b/packaging/docker/khiopspydev/Dockerfile.ubuntu
@@ -40,6 +40,8 @@ RUN true \
 # set up all the supported Python environments under conda (for the unit tests)
 # relying on a variable containing all the versions
 ARG PYTHON_VERSIONS
+ARG KHIOPS_GCS_DRIVER_REVISION
+ARG KHIOPS_S3_DRIVER_REVISION
 RUN true \
   && export CONDA="/root/miniforge3/bin/conda" \
   && /bin/bash -c 'for version in ${PYTHON_VERSIONS}; \
@@ -49,7 +51,12 @@ RUN true \
         do \
             $CONDA create -y -n $CONDA_ENV python=${version}; \
         done; \
+        # khiops core \
         $CONDA install -y -n py${version}_conda -c khiops-dev khiops-core=$(echo ${KHIOPS_REVISION} | tr -d "-") ; \
+        # remote files drivers installed in the conda environment \
+        $CONDA install -y -n py${version}_conda -c khiops \
+        khiops-driver-s3=${KHIOPS_S3_DRIVER_REVISION} \
+        khiops-driver-gcs=${KHIOPS_GCS_DRIVER_REVISION}; \
     done' \
   && true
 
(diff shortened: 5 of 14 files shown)


tramora:
@popescu-v & @folmos-at-orange : a detail we forgot the S3 & GCS drivers are not released publicly and thus not available in the dev image. When trying to access to the remote files during the tests khiops complains
```
error : there is no driver available for the URI scheme 'gs://'
error : there is no driver available for the URI scheme 's3://'
```

popescu-v:
> @popescu-v & @folmos-at-orange : a detail we forgot the S3 & GCS drivers are not released publicly and thus not available in the dev image. When trying to access to the remote files during the tests khiops complains
> 
> ```
> error : there is no driver available for the URI scheme 'gs://'
> error : there is no driver available for the URI scheme 's3://'
> ```

Yes, AFAIU we need to wait for the drivers to be publicly released before the tests can work.

.github/workflows/dev-docker.yml
```diff
@@ -5,6 +5,8 @@ env:
   DEFAULT_IMAGE_INCREMENT: 0
   DEFAULT_SERVER_REVISION: main
   DEFAULT_PYTHON_VERSIONS: 3.8 3.9 3.10 3.11 3.12
+  DEFAULT_KHIOPS_GCS_DRIVER_VERSION: 0.0.10
+  DEFAULT_KHIOPS_S3_DRIVER_VERSION: 0.0.12
```
popescu-v, line 9: I would use `_REVISION` instead of `_VERSION` to be consistent with `DEFAULT_KHIOPS_REVISION` and `DEFAULT_SERVER_REVISION`.

.github/workflows/dev-docker.yml
```diff
@@ -34,6 +36,14 @@ on:
         type: string
         default: main
         description: Khiops Server Revision
+      khiops-gcs-driver-version:
+        type: string
+        default: 0.0.10
+        description: Version of the driver used for remote files on Google Cloud Storage (GCS)
+      khiops-s3-driver-version:
```
popescu-v, line 43: I would use `-revision` instead of `-version` for consistency.

.github/workflows/unit-and-integration-tests.yml
```diff
@@ -1,5 +1,5 @@
 ---
-name: Unit Tests
+name: Unit & Integration Tests
```
popescu-v, line 2: I would go for simply "Tests" and `tests.yml`: shorter, more concise and allowing us more freedom to add different kinds of tests as well (non-regression, functional - actually, some Python "unit tests" are rather functional,  etc).

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -10,15 +10,14 @@ ARG KHIOPS_REVISION
...
   # Get Linux distribution codename \
   && if [ -f /etc/os-release ]; then . /etc/os-release; fi \
   # Obtain the Khiops native package \
   && KHIOPS_PKG_FILE=$KHIOPS_REVISION/khiops-core-openmpi_$KHIOPS_REVISION-1-$VERSION_CODENAME.amd64.deb \
   && wget -O KHIOPS_CORE.deb "https://github.com/KhiopsML/khiops/releases/download/${KHIOPS_PKG_FILE}" \
-  # Install the Khiops native package \
-  && dpkg -i --force-all KHIOPS_CORE.deb \
-  && apt-get -f -y install \
+  # Install the Khiops native package : do not break immediately if dpkg fails because apt will install the missing dependencies \
+  && (dpkg -i --force-all KHIOPS_CORE.deb || true) && apt-get -f -y install \
```
popescu-v, line 20: Split in two lines on ` && `.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -48,9 +47,37 @@ done'
...
 
+# new gcs & s3 drivers (written in C++)
+ARG KHIOPS_GCS_DRIVER_VERSION
+ARG KHIOPS_S3_DRIVER_VERSION
+RUN true \
+    && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_VERSION}/khiops-gcs_${KHIOPS_GCS_DRIVER_VERSION}.deb \
+    && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_VERSION}/khiops-s3_${KHIOPS_S3_DRIVER_VERSION}.deb \
+    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb || true) && apt-get -f -y install \
+    && rm -f khiops-gcs.deb khiops-s3.deb \
+    && true
```
popescu-v, line 62: s/`_VERSION`/`_REVISION`/ (see relevant comment above).

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -48,9 +47,37 @@ done'
...
+RUN chmod +x /scripts/*.sh && \
   useradd -rm -d /home/ubuntu -s /bin/bash -g root -u 1000 ubuntu
 
+# new gcs & s3 drivers (written in C++)
+ARG KHIOPS_GCS_DRIVER_VERSION
+ARG KHIOPS_S3_DRIVER_VERSION
+RUN true \
+    && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_VERSION}/khiops-gcs_${KHIOPS_GCS_DRIVER_VERSION}.deb \
+    && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_VERSION}/khiops-s3_${KHIOPS_S3_DRIVER_VERSION}.deb \
+    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb || true) && apt-get -f -y install \
```
popescu-v, line 60: Split in two lines on ` && `: thus each different command has its own line, which is consistent with the rest of the file.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -48,9 +47,37 @@ done'
...
+COPY ./run_fake_remote_file_servers.sh /scripts/run_fake_remote_file_servers.sh
+RUN chmod +x /scripts/*.sh && \
   useradd -rm -d /home/ubuntu -s /bin/bash -g root -u 1000 ubuntu
 
+# new gcs & s3 drivers (written in C++)
+ARG KHIOPS_GCS_DRIVER_VERSION
+ARG KHIOPS_S3_DRIVER_VERSION
+RUN true \
+    && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_VERSION}/khiops-gcs_${KHIOPS_GCS_DRIVER_VERSION}.deb \
+    && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_VERSION}/khiops-s3_${KHIOPS_S3_DRIVER_VERSION}.deb \
```
popescu-v, line 59: s/`_VERSION`/`_REVISION`/ (see relevant comment above).

packaging/docker/khiopspydev/run_fake_remote_file_servers.sh
```diff
@@ -0,0 +1,25 @@
...
+  -scheme http \
+  -public-host localhost &
+
+# File server for S3 (runs in background)
+# WARNING :
+# -r : exposes pre-provisioned files : the direct child folders will be the bucket names
+#      these files were uploaded once because fake-s3 creates metadata
+nohup /usr/local/bin/fakes3 \
+  -r "${ROOT_FOLDER}"/tests/resources/remote-access \
+  -p 4569 &
```
popescu-v, line 25: IMHO the port should not be hard-coded, but instead be passed as a variable or extracted from `AWS_ENDPOINT_URL`.

tests/resources/remote-access/s3-bucket/khiops-cicd/samples/Iris/Iris.kdic/.fakes3_metadataFFF/metadata
```diff
@@ -0,0 +1,8 @@
+---
```
popescu-v, line 1: Do we need to commit these metadata files?

tests/test_remote_access.py
```diff
@@ -27,18 +27,26 @@
 
 
 def s3_config_exists():
+    # instead of the indirect env vars
+    # `AWS_SHARED_CREDENTIALS_FILE` and `AWS_CONFIG_FILE` that references files
```
popescu-v, line 31: AFAIK, the advantage of using an `AWS_SHARED_CREDENTIALS_FILE` is to centralize the "sensitive" information into a file that can be protected through access control so that only dedicated user (in dedicated group) can read it. And besides, the current modification obliges to expose the "sensitive" credentials a such ("in clear") in the environment. I'm not sure this is good practice.

Hence, I would maintain the file-based configuration:
- create the file on the Docker container
- add relevant credentials within
- protect the file through `chmod`.

.github/workflows/dev-docker.yml
tramora: Variable name changed

.github/workflows/dev-docker.yml
tramora: Variable name changed

.github/workflows/unit-and-integration-tests.yml
tramora: Label changed and yaml file renamed

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: reformatting done

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -48,9 +47,37 @@ done'
 
 RUN mkdir -p /scripts
 COPY ./run_service.sh /scripts/run_service.sh
-RUN chmod +x /scripts/run_service.sh && \
+COPY ./run_fake_remote_file_servers.sh /scripts/run_fake_remote_file_servers.sh
+RUN chmod +x /scripts/*.sh && \
```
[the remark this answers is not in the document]
tramora: precision done

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: reformatting done

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: Variable name changed

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -48,9 +47,37 @@ done'
...
+
 FROM khiopsdev AS base
 COPY --from=server /service /usr/bin/service
+
+# GCS fake file server (only in the ubuntu container)
+COPY --from=gcs-server /bin/fake-gcs-server /bin/fake-gcs-server
+EXPOSE 4443
+
+# S3 fake file server (only in the ubuntu container)
+# Do not use the latest fakes3 version because starting from 1.3.1 a licence is required
```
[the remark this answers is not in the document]
tramora: The pinned version of fakes3 covers all the needs for a remote access with the possibility of pre-provisionning the server with test files before it start.

Accordingly to the documentation s3rver does not provide this feature (yet)

packaging/docker/khiopspydev/run_fake_remote_file_servers.sh
tramora: port number extracted from the URL

tests/resources/remote-access/s3-bucket/khiops-cicd/samples/Iris/Iris.kdic/.fakes3_metadataFFF/metadata
tramora: Yes unfortunately, fakes3 needs these files otherwise the preprovisioned files are ignored

packaging/docker/khiopspydev/Dockerfile.ubuntu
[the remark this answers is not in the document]
popescu-v: OK, then, to mitigate further maintenance issue, I would state this in a TODO comment; something like:
```
# TODO: Switch to https://github.com/jamhall/s3rver (MIT-licensed) as soon as it supports pre-provisioning the server with fixtures.
```
?

tests/resources/remote-access/s3-bucket/khiops-cicd/samples/Iris/Iris.kdic/.fakes3_metadataFFF/metadata
popescu-v: OK, then I would put all these metadata files in a separate commit whose commit message states precisely this. Thus, this could be reverted more easily once the switch to s3ver (or another alternative) is done.

.github/workflows/tests.yml
```diff
@@ -314,3 +314,82 @@ jobs:
 
           # Execute Khiops sample (train and deploy model), which uses MPI
           python -m khiops.samples.samples -i deploy_model -e
+  test-access-to-remote-files:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/khiopsml/khiops-python/khiopspydev-ubuntu22.04:10.2.3-b.5.s3-gcs-remote-files
```
popescu-v, line 320: IMHO, all tests should be unified in a single job which uses this docker file tag. IMHO, we can update the default value for the tag, as well as for the inputs (on this branch). This was also done for other PRs (like switching to conda-forge etc).

But this update (in all impacted workflows, including pip or api-docs !) should be in a separate commit IMHO so that we identify it easily.

packaging/docker/khiopspydev/run_fake_remote_file_servers.sh
```diff
@@ -0,0 +1,26 @@
...
+  -public-host localhost &
+
+# File server for S3 (runs in background)
+# WARNING :
+# -r : exposes pre-provisioned files : the direct child folders will be the bucket names
+#      these files were uploaded once because fake-s3 creates metadata
+PORT_NUMBER=${AWS_ENDPOINT_URL##*:}
+nohup /usr/local/bin/fakes3 \
+  -r "${ROOT_FOLDER}"/tests/resources/remote-access \
+  -p "${PORT_NUMBER:-4569}" &
```
popescu-v, line 26: Should only be `-p $PORT_NUMBER` IMHO.

tests/resources/remote-access/gcs-bucket/khiops-cicd/samples/Iris/Iris.txt
folmos-at-orange, line 1: We really should avoid adding data files to the repository. These can be retrieved with the `kh-download-samples` tool.

folmos-at-orange (changes requested):

tests/resources/remote-access/s3-bucket/khiops-cicd/samples/Iris/Iris.kdic/.fakes3_metadataFFF/metadata
popescu-v: If `kh-download-datasets` is used to download the samples, then there is no need of pre-provisioning the files with the storage. Hence, there is no need for storing the metadata files anymore.

tests/resources/remote-access/gcs-bucket/khiops-cicd/samples/Iris/Iris.txt
folmos-at-orange: I saw the files removed. I resolve this.

popescu-v:
> There is an issue remaining that breaks the tests and that I cannot fix right now.
> 
> It seems that khiops under the conda env cannot use the remote files drivers installed system wide. They cannot be installed via conda as there is no conda release yet.

This is true. Hence, IMHO the way to go for the current PR is to test the remote access only in the native case (Conda environments for test, but with the native `khiops-core` and drivers packages installed).

Then, as soon as the Conda packages for the drivers are available, a new PR should activate these integration tests in full Conda environments.

.github/workflows/unit-and-integration-tests.yml
```diff
@@ -314,3 +314,82 @@ jobs:
 
           # Execute Khiops sample (train and deploy model), which uses MPI
           python -m khiops.samples.samples -i deploy_model -e
+  test-access-to-remote-files:
```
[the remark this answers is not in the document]
tramora: No duplication anymore

packaging/docker/khiopspydev/Dockerfile.ubuntu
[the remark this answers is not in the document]
tramora: I started evaluating jamhall/s3rver but unfortunatly it is not fully compatible with our basic usage.
I ended in putting this comment
```
# if fakes3 is no longer compatible think about switching to an alternative and fully compatible server
```

tests/resources/remote-access/s3-bucket/khiops-cicd/samples/Iris/Iris.kdic/.fakes3_metadataFFF/metadata
tramora: no pre-provisioning is done any longer, every test is set up with previously downloaded samples

tests/test_remote_access.py
tramora: The alternative AWS_* env variables are replaced by a configuration and credentials files referenced by the corresponding variables (AWS_CONFIG_FILE and AWS_SHARED_CREDENTIALS_FILE)

tests/test_remote_access.py
```diff
@@ -27,18 +27,26 @@
...
         and "S3_BUCKET_NAME" in os.environ
     )
 
 
 def gcs_config_exists():
-    return (
-        "GOOGLE_APPLICATION_CREDENTIALS" in os.environ
-        and "GCS_BUCKET_NAME" in os.environ
-    )
+    # for the integration tests
```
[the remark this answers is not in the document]
tramora: Rollback to the check to the pair of env variables GOOGLE_APPLICATION_CREDENTIALS and GCS_BUCKET_NAME

.github/workflows/tests.yml

packaging/docker/khiopspydev/run_fake_remote_file_servers.sh
tramora: Fixed

tests/test_remote_access.py
```diff
@@ -79,17 +119,52 @@ def skip_if_no_config(self):
...
+            # Cleanup the output dir (the files within and the folder)
+            if hasattr(self, "folder_name_to_clean_in_teardown"):
+                for filename in fs.list_dir(self.folder_name_to_clean_in_teardown):
+                    fs.remove(
+                        fs.get_child_path(
+                            self.folder_name_to_clean_in_teardown, filename
+                        )
+                    )
+                fs.remove(self.folder_name_to_clean_in_teardown)
+
```
folmos-at-orange, line 156: To avoid using `hasattr` and setting up your own temporary urls, instead of a per test setup/tear-down, I'd use one for the whole class. 

- In the `setupClass` method you simply set the env var `KHIOPS_TMP_DIR` with a fresh runner.
- In the `tearDownClass` method you clean up everything and restore the original temporary directory and runner.

Something like this example
```python
import unittest
import os
import subprocess
from khiops import core as kh
from khiops.core.internals.runner import KhiopsLocalRunner


class TestWithAltRunner(unittest.TestCase):
    @classmethod
    def setUpClass(cls):
        # Save the original runner and temp dir
        cls.original_runner = kh.get_runner()
        cls.original_khiops_temp_dir = os.environ.get("KHIOPS_TMP_DIR")

        # Set the test suite temp dir
        os.environ["KHIOPS_TMP_DIR"] = "/tmp/mytmpdir/" # replace here with the URL
        cls.runner = KhiopsLocalRunner()
    @classmethod
    def tearDownClass(cls):
        # Restore the original runner and temp dir
        kh.set_runner(cls.original_runner)
        if cls.original_khiops_temp_dir is None:
            del os.environ["KHIOPS_TMP_DIR"]
        else:
            os.environ["KHIOPS_TMP_DIR"] = cls.original_khiops_temp_dir

        # Clean the whole alternative temporary file
        # ...

    def test(self):
        print("\n======= Current test runner =========")
        kh.get_runner().print_status()
```

folmos-at-orange (changes requested):

tests/test_remote_access.py
```diff
@@ -341,6 +441,12 @@ def tearDownClass(cls):
     def config_exists(self):
         return docker_runner_config_exists()
 
+    def should_skip_in_a_conda_env(self):
+        """
+        Tests using a docker runner should never be skipped even in a conda env
+        """
```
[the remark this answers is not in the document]
tramora: Fixed

tests/test_remote_access.py
```diff
@@ -79,17 +119,52 @@ def skip_if_no_config(self):
                     "has no configuration available"
                 )
 
+        @staticmethod
+        def is_in_a_conda_env():
+            # Debug :
+            # print(
+            #     f"{self.remote_access_test_case()} : "
+            #     f"Variables available when testing if in a CONDA env : {os.environ}"
+            # )
```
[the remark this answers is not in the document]
tramora: The code was left in comment if later debug is needed. Comment removed

tests/test_remote_access.py
```diff
@@ -69,6 +99,16 @@ def remote_access_test_case(self):
             """To be overriden by descendants"""
             return ""
 
+        def should_skip_in_a_conda_env(self):
+            """
+            To be overriden by descendants
+
+            Temporary skip the S3 and GCS tests under a conda environment
+            until the drivers for remote files access
+            are released as conda packages
```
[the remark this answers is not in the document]
tramora: Typo fixed

tests/test_remote_access.py
```diff
@@ -54,6 +59,31 @@ class KhiopsRemoteAccessTestsContainer:
...
+                "SpliceJunction/SpliceJunctionDNA.txt",
+                "SpliceJunction/SpliceJunction.kdic",
+            ):
+                fs.copy_from_local(
+                    f"{proto}://{bucket_name}/khiops-cicd/samples/" + file,
+                    os.path.join(kh.get_samples_dir(), file),
+                )
+                # symmetric call to ensure the upload was OK
+                fs.copy_to_local(
+                    f"{proto}://{bucket_name}/khiops-cicd/samples/" + file, "/tmp/dummy"
```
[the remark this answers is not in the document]
tramora: Fixed

.github/workflows/tests.yml
```diff
@@ -112,7 +112,30 @@ jobs:
...
         if: github.ref != 'dev' && github.ref != 'main' && ! inputs.run-long-tests
         run: echo "UNITTEST_ONLY_SHORT_TESTS=true" >> "$GITHUB_ENV"
-      - name: Run Unit Tests
+      - name: Prepare Integration Tests on remote files
+        env:
+          AWS_ENDPOINT_URL: http://localhost:4569
+        shell: bash
+        run: |
+          # for S3 only
+          echo "HOME=${HOME}"
```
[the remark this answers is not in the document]
tramora: The command was put for diagnosis purpose : the HOME env variable was not expanded everywhere and I needed to see its value to copy it. Command removed

.github/workflows/tests.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       # because the `env` context is only accessible at the step level;
       # hence, it is hard-coded
       image: |-
-        ghcr.io/khiopsml/khiops-python/khiopspydev-ubuntu22.04:${{ inputs.image-tag || 'latest' }}
+        ghcr.io/khiopsml/khiops-python/khiopspydev-ubuntu22.04:${{ inputs.image-tag || '10.2.3.0.s3-gcs-remote-files' }}
```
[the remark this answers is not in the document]
tramora: I may have misunderstood the process agreed with @popescu-v.
As long as the PR is in review,  the specific image build with this branch is not set yet as the latest.
Just before the merge then I can change to the latest.

.github/workflows/dev-docker.yml
```diff
@@ -34,6 +36,14 @@ on:
         type: string
         default: main
         description: Khiops Server Revision
+      khiops-gcs-driver-revision:
+        type: string
+        default: 0.0.10
+        description: Version of the driver used for remote files on Google Cloud Storage (GCS)
```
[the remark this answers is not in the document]
tramora: Rewording done

khiops/core/internals/filesystems.py
```diff
@@ -668,29 +668,26 @@ def remove(self):
...
-                f"S3 copy_from_local failed {self.uri} with code {status_code}: "
-                + json.dumps(response)
+        # boto3.s3.transfer.upload_file is not expected to return an HTTP response
+        # but instead raises a S3UploadFailedError exception in case of error
+        try:
+            self.s3_client.Bucket(self.uri_info.netloc).upload_file(
+                local_path, self.uri_info.path[1:]
             )
+        # normalize the raised exception
+        except Exception as e:
```
popescu-v, line 678: I would also restrict the "catch" here to `S3UploadFailedError`:
```python
except S3UploadFailedError as exc:
    raise RuntimeError(...) from exc
```
Thus, wenever there is a different (unexpected) exception, it would not be handled and whence would be propagated to the caller code so that we know about it (e.g. from inspecting the CI logs).

packaging/conda/meta.yaml
```diff
@@ -30,8 +30,11 @@ requirements:
     - pandas >=0.25.3
     - scikit-learn >=0.22.2
   run_constrained:
-    - boto3 >=1.17.39
+    # do not necessary use the latest version
+    # to avoid undesired breaking changes
+    - boto3 >=1.17.39,<=1.35.69
     - google-cloud-storage >=1.37.0
+    - pyopenssl>=24.0.0
```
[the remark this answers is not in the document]
popescu-v: Do we need `pyopenssl` to impose a specific lower bound of the version?

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -54,9 +54,43 @@ RUN true \
...
+
+# S3 fake file server (only in the ubuntu container)
+# Do not use the latest fakes3 version because starting from 1.3 a licence is required
+# if fakes3 is no longer compatible think about switching to an alternative and fully compatible server
+# (https://github.com/jamhall/s3rver:v3.7.1 is not yet for example)
+RUN gem install fakes3:1.2.1 sorted_set
+# Avoid resolving a fake s3-bucket.localhost hostname
+# Alternate builders (buildx via moby buildkit) mount /etc/hosts read-only, the following command will fail
+# echo "127.0.0.1 s3-bucket.localhost" >> /etc/hosts
+# You will have to add the `add-hosts` input instead (https://github.com/docker/build-push-action/#inputs)
```
popescu-v, line 92: Why can't we do this directly here? This Docker file is only useful in the CI context anyway.

tests/test_remote_access.py
```diff
@@ -27,6 +27,11 @@
 
 
 def s3_config_exists():
+    # The alternate env vars...
+    # (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_ENDPOINT_URL`)
+    # plus the bucket name in `S3_BUCKET_NAME`
+    # are also usable instead of the credentials and config files
+    # but are regarded unsecure
```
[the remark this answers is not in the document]
popescu-v: I would replace the last phrase "However, ..." with emphasis on the improved security of the alternative approach, like:

```bash
# However, storing the credentials in config files is more secure, because these can be protected by access policies.
```

tests/test_remote_access.py
```diff
@@ -79,17 +118,47 @@ def skip_if_no_config(self):
                     "has no configuration available"
                 )
 
+        @staticmethod
+        def is_in_a_conda_env():
+            # CONDA_DEFAULT_ENV and CONDA_PROMPT_MODIFIER
```
popescu-v, line 123: Rather `CONDA_PREFIX` than `CONDA_DEFAULT_ENV` (the former gives full path to the env).

But this would not work in a "conda-based" environment anyway (i.e. khiops is installed in a Conda env, but code is run from outside the Conda env (i.e. `CONDA_*` vars are not set: look for "conda-based" in runner.py).

Hence, we need to find a solution for this, because we don't want to use native drivers with Conda-installed khiops packages.

A solution might be to rely on `kh.get_runner().print_status()` and to look-up "install type" in the output of the function. `is_in_a_conda_env()` would return `True` if install type is either "conda" or "conda-based".

Something like:

```python
# At the beginning (if not already done)
from khiops import core as kh
#...

@staticmethod
def is_in_a_conda_env():
    status_message = kh.get_runner()._build_status_message()[0]
    for status_line in status_message.splitlines():
        if "install type" in status_line:
            install_type = status_line.split(":").strip()
            break
     return install_type in ("conda", "conda-based")
```

tests/test_remote_access.py
```diff
@@ -79,17 +118,47 @@ def skip_if_no_config(self):
...
+            # are the only usable env variables that let us know we are in a conda env
+            conda_default_env = os.environ.get("CONDA_DEFAULT_ENV")
+            return conda_default_env and conda_default_env.endswith("_conda")
```
popescu-v, line 126: The `.endswith` leaks CI details in the tests (as the `_conda` env names for Conda-installed `khiops-core` is a CI detail).

The proposal from the previous comment avoids this.

tests/test_remote_access.py
```diff
@@ -79,17 +118,47 @@ def skip_if_no_config(self):
...
+
         def setUp(self):
             self.skip_if_no_config()
+            if self.is_in_a_conda_env() and self.should_skip_in_a_conda_env():
+                # TODO : Temporary skip the test under a conda environment
```
popescu-v, line 131: Rephrase TODO comment to something like
`# TODO: Temporary skip the test for Conda-installed khiops-core`

tests/test_remote_access.py
```diff
@@ -204,9 +282,17 @@ def setUpClass(cls):
...
             bucket_name = os.environ["S3_BUCKET_NAME"]
-            runner.samples_dir = f"s3://{bucket_name}/project/khiops-cicd/samples"
-            cls._khiops_temp_dir = f"s3://{bucket_name}/project/khiops-cicd/tmp"
-            runner.root_temp_dir = f"s3://{bucket_name}/project/khiops-cicd/tmp"
+
+            cls.init_remote_bucket(bucket_name=bucket_name, proto="s3")
+
+            runner.samples_dir = f"s3://{bucket_name}/khiops-cicd/samples"
+            resources_directory = KhiopsTestHelper.get_resources_dir()
+            # WARNING : khiops temp files cannot be remote
```
popescu-v, line 290: Add empty line before the comment (to have paragraphs).

tests/test_remote_access.py
```diff
@@ -204,9 +282,17 @@ def setUpClass(cls):
...
+            cls._khiops_temp_dir = f"{resources_directory}/tmp/khiops-cicd"
+            # whereas root_temp_dir
```
popescu-v, line 292: Delete "whereas".

Add empty line before the beginning of the comment (to have the commented block as a paragraph).

tests/test_remote_access.py
```diff
@@ -232,8 +318,16 @@ def setUpClass(cls):
         if gcs_config_exists():
             runner = kh.get_runner()
             bucket_name = os.environ["GCS_BUCKET_NAME"]
+
+            cls.init_remote_bucket(bucket_name=bucket_name, proto="gs")
+
             runner.samples_dir = f"gs://{bucket_name}/khiops-cicd/samples"
-            cls._khiops_temp_dir = f"gs://{bucket_name}/khiops-cicd/tmp"
+            resources_directory = KhiopsTestHelper.get_resources_dir()
+            # WARNING : khiops temp files cannot be remote
```
popescu-v, line 326: Add empty line before comment.

tests/test_remote_access.py
```diff
@@ -232,8 +318,16 @@ def setUpClass(cls):
...
+            cls._khiops_temp_dir = f"{resources_directory}/tmp/khiops-cicd"
+            # whereas root_temp_dir
```
popescu-v, line 328: Delete "whereas". Add empty line before the first line of the comment.

popescu-v (changes requested):
See the comments.

setup.py
```diff
@@ -49,7 +49,12 @@
...
         cmdclass=versioneer.get_cmdclass(),
         extras_require={
-            "s3": ["boto3>=1.17.39"],
+            "s3": [
+                # do not necessary use the latest version
+                # to avoid undesired breaking changes
+                "boto3>=1.17.39,<=1.35.69"
+            ],
             "gcs": ["google-cloud-storage>=1.37.0"],
+            "pyopenssl": ["pyopenssl>=24.0.0"],
```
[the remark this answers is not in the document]
folmos-at-orange: Then the dependency should impose its restriction on this package. Unless there are serious problems we should remove this.

.github/workflows/tests.yml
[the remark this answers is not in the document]
folmos-at-orange: Ok. To fix beforem merging then.

setup.py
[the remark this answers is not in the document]
popescu-v: > Then the dependency should impose its restriction on this package. Unless there are serious problems we should remove this.

@tramora : Is there any specific version pinning issue with this package?

packaging/docker/khiopspydev/Dockerfile.ubuntu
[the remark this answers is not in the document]
popescu-v: Ok.

.github/workflows/tests.yml
popescu-v: Ok.

tests/test_remote_access.py
```diff
@@ -204,9 +300,18 @@ def setUpClass(cls):
...
-            runner.root_temp_dir = f"s3://{bucket_name}/project/khiops-cicd/tmp"
+
+            cls.init_remote_bucket(bucket_name=bucket_name, proto="s3")
+
+            runner.samples_dir = f"s3://{bucket_name}/khiops-cicd/samples"
+            resources_directory = KhiopsTestHelper.get_resources_dir()
+
+            # WARNING : khiops temp files cannot be remote
+            cls._khiops_temp_dir = f"{resources_directory}/tmp/khiops-cicd"
+            # root_temp_dir
```
popescu-v, line 311: Add empty line before comment to make it a paragraph.

tests/test_remote_access.py
```diff
@@ -232,8 +337,17 @@ def setUpClass(cls):
...
+
+            cls.init_remote_bucket(bucket_name=bucket_name, proto="gs")
+
             runner.samples_dir = f"gs://{bucket_name}/khiops-cicd/samples"
-            cls._khiops_temp_dir = f"gs://{bucket_name}/khiops-cicd/tmp"
+            resources_directory = KhiopsTestHelper.get_resources_dir()
+
+            # WARNING : khiops temp files cannot be remote
+            cls._khiops_temp_dir = f"{resources_directory}/tmp/khiops-cicd"
+            # root_temp_dir
```
popescu-v, line 348: Add empty line before comment to make it a paragraph.

tests/test_remote_access.py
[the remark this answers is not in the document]
folmos-at-orange: I still see `Temporary` (-> `Temporarily`)

tests/test_remote_access.py
folmos-at-orange: I created the issue https://github.com/KhiopsML/khiops-python/issues/328.

**You may resolve this once you have read this message**

setup.py
```diff
@@ -49,7 +49,13 @@
...
+            "s3": [
+                # do not necessary use the latest version
+                # to avoid undesired breaking changes
+                "boto3>=1.17.39,<=1.35.69"
+            ],
             "gcs": ["google-cloud-storage>=1.37.0"],
+            # an open issue on boto3 (https://github.com/boto/boto3/issues/3585) forces a min version of pyopenssl
+            "pyopenssl": ["pyopenssl>=24.0.0,<25.0.0"],
         },
     )
```
folmos-at-orange, line 61: The `pyopenssl` dependency should go in the `s3` list, as it only affects the `boto3` extra.

tests/test_remote_access.py
```diff
@@ -79,17 +123,67 @@ def skip_if_no_config(self):
                     "has no configuration available"
                 )
 
+        @staticmethod
+        def is_in_a_conda_env():
+            """Detects whether this is run from a Conda environment
+            and khiops-core is installed in the same env
+            """
```
folmos-at-orange, line 130: Titles should fit the line.

The title is the minimal summary. If there are more details put in in the body.

tests/test_remote_access.py
```diff
@@ -341,6 +463,12 @@ def tearDownClass(cls):
     def config_exists(self):
         return docker_runner_config_exists()
 
+    def should_skip_in_a_conda_env(self):
+        """Tests using a docker runner should never be skipped
+        even in a conda env
+        """
```
folmos-at-orange, line 469: I think this is `#` comment, not a docstring.

folmos-at-orange (changes requested):

run_fake_remote_file_servers.sh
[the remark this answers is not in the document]
popescu-v: I also fail to see the need for this top-level symlink to the real file which is in the `packaging/docker/khiopspydev` directory, and which is copied, in the `Dockerfile`, to the in-container `/scripts` directory.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -54,9 +55,43 @@ RUN true \
 
 RUN mkdir -p /scripts
 COPY ./run_service.sh /scripts/run_service.sh
-RUN chmod +x /scripts/run_service.sh && \
+COPY ./run_fake_remote_file_servers.sh /scripts/run_fake_remote_file_servers.sh
```
popescu-v, line 58: Conflate with line n° 57 into ( the trailing `/` for the destination directory is needed when copying several files IIRC):

```sh
COPY ./run_service.sh ./run_fake_remote_servers.sh /scripts/
```

tramora: layers optimization done

tests/test_remote_access.py
tramora: Fixed

tests/test_remote_access.py
tramora: Fixed

setup.py
tramora: Fixed

.github/workflows/tests.yml
```diff
@@ -112,7 +112,30 @@ jobs:
       - name: Prepare Unit Tests Environment
         if: github.ref != 'dev' && github.ref != 'main' && ! inputs.run-long-tests
         run: echo "UNITTEST_ONLY_SHORT_TESTS=true" >> "$GITHUB_ENV"
-      - name: Run Unit Tests
+      - name: Prepare Integration Tests on remote files
+        env:
+          AWS_ENDPOINT_URL: http://localhost:4569
+          GITHUB_WORKSPACE: ${{ github.workspace }}
```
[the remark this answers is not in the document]

tests/test_remote_access.py
```diff
@@ -79,17 +123,67 @@ def skip_if_no_config(self):
...
+
+            if not isinstance(kh.get_runner(), KhiopsLocalRunner):
+                return False
+
+            # Get path to the Khiops executable
+            khiops_path = kh.get_runner()._khiops_path
+
+            # If $(dirname khiops_path) is identical to $CONDA_PREFIX/bin,
+            # then return True
+            print(
```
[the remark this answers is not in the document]
tramora: Fixed

tests/test_remote_access.py
tramora: Fixed

tests/test_remote_access.py
tramora: Fixed

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: Unfortunately as stated in the comment, the modification /etc/hosts is not possible here as buildx will mount the / filesystem readonly.

folmos-at-orange (approved):

packaging/docker/khiopspydev/Dockerfile.ubuntu
popescu-v: Ok.

popescu-v (changes requested):
Address https://github.com/KhiopsML/khiops-python/pull/210#discussion_r1891415778.

Shorten commit message of https://github.com/KhiopsML/khiops-python/pull/210/commits/2dbb34e66dfa9edd41d02f89c9fbb7573f752323 to "Update copy_from_local and copy_to_local for S3".

tramora:
> Address [#210 (comment)](https://github.com/KhiopsML/khiops-python/pull/210#discussion_r1891415778).
> 
> Shorten commit message of [2dbb34e](https://github.com/KhiopsML/khiops-python/commit/2dbb34e66dfa9edd41d02f89c9fbb7573f752323) to something like "Update copy_from_local and copy_to_local for S3".

Both issues addressed

.github/workflows/tests.yml
```diff
@@ -150,12 +160,10 @@ jobs:
...
-          GCS_BUCKET_NAME: gcs-bucket
-          # we take advantage of the built-in `STORAGE_EMULATOR_HOST` env variable
-          # that every GCS client can read and lets us use a local fake file server
-          STORAGE_EMULATOR_HOST: http://localhost:4443
-          # even in GCS emulation mode, the credentials file must exist
-          GOOGLE_APPLICATION_CREDENTIALS: ${{ github.workspace }}/google-credentials.json
+          GCS_BUCKET_NAME: khiops-integtest-data
+          # Remark : in case we use a fake GCS server,
+          # the built-in `STORAGE_EMULATOR_HOST` and `CLOUD_STORAGE_EMULATOR_ENDPOINT` env variables
+          # that every GCS client can read must also be set
```
popescu-v, line 166: For the time being, we cannot use any known available fake GCS server, as they are not supported. Hence, I would remove these comment lines.

.github/workflows/tests.yml
```diff
@@ -133,10 +133,20 @@ jobs:
...
           /scripts/run_fake_remote_file_servers.sh .  # launch the servers in the background
+      - name: Authenticate to GCP using "Workload Identity Federation"
+        # For integration tests on GCS we use a real Google account
+        # Retrieve the Google credentials through "Workload Identity Federation"
+        # see https://github.com/google-github-actions/auth?tab=readme-ov-file#workload-identity-federation-through-a-service-account
+        uses: google-github-actions/auth@v2
+        with:
+          service_account: wif-test@ino-ols-khiops-aas-prd.iam.gserviceaccount.com
+          workload_identity_provider: projects/345517370475/locations/global/workloadIdentityPools/github/providers/github
+          token_format: access_token
```
popescu-v, line 145: Why is this line needed here, whereas in https://github.com/KhiopsML/khiopsdriver-gcs/blob/main/.github/workflows/pack-conda.yml#L112 it is not?

tramora: Requesting an OAuth2 https://github.com/google-github-actions/auth?tab=readme-ov-file#inputs-generating-oauth-20-access-tokens had the border advantage of failing at this authentication step if the google service account is misconfigured. The feeback is a few minutes later (during the tests) otherwise but OK to remove this call.

.github/workflows/tests.yml
tramora: The comments were removed

.github/workflows/tests.yml
```diff
@@ -124,6 +158,16 @@ jobs:
           rmaps_base_oversubscribe: true
           # Oversubscribe for MPI > 4.x
           OMPI_MCA_rmaps_base_oversubscribe: true
+          # for the tests with GCS
+          GCS_BUCKET_NAME: data-test-khiops-driver-gcs/khiops_data
+          GCS_DRIVER_LOGLEVEL: info  # set to debug for diagnosis
+          S3_DRIVER_LOGLEVEL: info  # set to debug for diagnosis
+          # for the tests with S3
```
popescu-v, line 165: Move this comment one line above.

.github/workflows/tests.yml
```diff
@@ -124,6 +158,16 @@ jobs:
...
+          S3_BUCKET_NAME: s3-bucket
+          AWS_SHARED_CREDENTIALS_FILE: ${{ github.workspace }}/.aws/credentials
+          AWS_CONFIG_FILE: ${{ github.workspace }}/.aws/configuration
+          # common var for tests with GCS & S3
```
popescu-v, line 169: Update comment to something like: `# Var for tests with S3` (it does not apply to GCS anymore AFAIU).

.github/workflows/tests.yml
popescu-v: Ok.

packaging/docker/khiopspydev/run_fake_remote_file_servers.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
popescu-v, line 1: Rename file to "run_fake_s3_remote_file_server.sh".

.github/workflows/tests.yml

.github/workflows/tests.yml
tramora: Fixed

tests/test_remote_access.py
```diff
@@ -79,17 +118,58 @@ def skip_if_no_config(self):
...
+
+            # If $(dirname khiops_path) is identical to $CONDA_PREFIX/bin,
+            # then return True
+            conda_prefix = os.environ.get("CONDA_PREFIX")
+            return conda_prefix is not None and os.path.join(
+                conda_prefix, "bin"
+            ) == os.path.dirname(khiops_path)
+
         def setUp(self):
             self.skip_if_no_config()
```
popescu-v, line 143: IMHO we should skip these tests unless a specific environment variable is set via a specific CI workflow input (similar to the long tests switching). Thusly, we make sure these tests are run on-demand, whence avoiding to use the GCS bucket on each commit on each MR.

tramora: reuse of the `skip_long_test` mechanism done (renamed to `skip_expensive_test`)

tests/test_helper.py
```diff
@@ -272,10 +272,10 @@ def get_with_subkey(dictionary, subkey):
         return values
 
     @staticmethod
-    def skip_long_test(test_case):
+    def skip_expensive_test(test_case):
         if "UNITTEST_ONLY_SHORT_TESTS" in os.environ:
```
popescu-v, line 276: I would rename this variable to something like `SKIP_EXPENSIVE_TESTS`, while keeping the same execution semantics.

.github/workflows/tests.yml
```diff
@@ -110,7 +110,7 @@ jobs:
             rm -rf khiops.egg-info
           done
       - name: Prepare Unit Tests Environment
-        if: github.ref != 'dev' && github.ref != 'main' && ! inputs.run-long-tests
+        if: github.ref != 'dev' && github.ref != 'main' && ! inputs.run-expensive-tests
         run: echo "UNITTEST_ONLY_SHORT_TESTS=true" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
tramora: done

tests/test_remote_access.py
```diff
@@ -453,8 +453,9 @@ def tearDownClass(cls):
 
         # Cleanup: remove directories created in `setUpClass`
         if docker_runner_config_exists():
-            shutil.rmtree(kh.get_runner().samples_dir)
-            shutil.rmtree(cls._khiops_temp_dir)
+            with suppress(FileNotFoundError):
```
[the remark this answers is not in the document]
tramora: done

popescu-v (approved):
LGTM.

However, https://github.com/KhiopsML/khiops-python/pull/210#discussion_r1907345903 needs to be taken into account before merging into `dev-v10`.

PR #218, opened 2024-07-31
209 openmpi path issue with khiops 1022 on debian like os

### LINKED ISSUE #209: OpenMPI path issue with khiops 10.2.2 on Debian-like OS
#### Description
When using khiops-python API, the following command line is eventually executed :

```
/bin/mpiexec --allow-run-as-root --quiet -n 14 /usr/bin/MODL -b -i train_predictor_XXX._kh -e khiops.log
```

With openmpi installed with this 10.2.2-1 version it fails :

```
--------------------------------------------------------------------------
A prefix was supplied to mpiexec that only contained slashes.

This is a fatal error; mpiexec will now abort.  No processes were launched.
--------------------------------------------------------------------------
```

#### Questions/Ideas

This issue is already identified on the openmpi project https://github.com/open-mpi/ompi/issues/5613.

To my understanding, the issue exists on systems (Arch, Ubuntu) where /bin/ is a symlink to /usr/bin. A kind of nasty loop occurs.

A fix on khiops-python was already committed once https://github.com/KhiopsML/khiops-python/commit/75d0bf70a621092e4db461072185ace4a9f35688 but unfortunately seems to have been reverted.

The symlink was followed and the real binary was reached.
```
/bin/mpiexec -> /etc/alternatives/mpiexec -> /usr/bin/mpiexec.openmpi    
```

I think we should put back this fix.

popescu-v:
closes #209

### DIFF
--- a/.github/workflows/pip.yml
+++ b/.github/workflows/pip.yml
@@ -93,8 +93,11 @@ jobs:
           kh-samples sklearn -i khiops_classifier -e
 
           # Test that the line containing "MPI command" also contains
-          # "mpiexec", which means that `mpiexec` has been found
-          kh-status | grep "MPI command" | grep -wq mpiexec
+          # an executable name under a /bin directory
+          # Note: this executable name can be different, depending on the MPI
+          # backend and OS; for instance, "orterun" for OpenMPI on Ubuntu Linux, but
+          # "mpiexec" for OpenMPI on Rocky Linux
+          kh-status | grep "MPI command" | grep -Ewq "(/.+?)/bin/.+"
   release:
     if: github.ref_type == 'tag'
     needs: [build, test]
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -99,6 +99,9 @@ jobs:
           OMPI_MCA_rmaps_base_oversubscribe: true
           PRTE_MCA_rmaps_default_mapping_policy: :oversubscribe
         run: |
+          # Make sure '/bin' is before '/usr/bin' in PATH
+          PATH=$(echo "/bin:"$PATH | sed 's#:/bin##')
+
           # This is needed so that the Git tag is parsed and the khiops-python
           # version is retrieved
           git config --global --add safe.directory $(realpath .)
@@ -177,6 +180,9 @@ jobs:
           # Force > 2 CPU cores to launch mpiexec
           KHIOPS_PROC_NUMBER: 4
         run: |-
+          # Make sure '/bin' is before '/usr/bin' in PATH
+          PATH=$(echo "/bin:"$PATH | sed 's#:/bin##')
+
           # Make sure MPI support is not loaded through env modules
           # Note: As Docker container's shell is non-interactive, environment
           # modules are currently not initializing the shell anyway
--- a/tests/test_dataset_class.py
+++ b/tests/test_dataset_class.py
@@ -565,13 +565,15 @@ class KhiopsConsistensyOfFilesAndDictionariesWithInputDataTests(unittest.TestCas
             target, features = line.split(b"\t")
             feature_row = np.zeros(100)
             for feature in features.strip().split(b" "):
-                feature_index, feature_value = feature.split(b":")
-                try:
-                    feature_value = float(feature_value)
-                # missing value, whence empty string
-                except ValueError:
-                    feature_value = 0.0
-                feature_row[int(feature_index) - 1] = feature_value
+                indexed_feature = feature.split(b":")
+
+                # Skip missing feature
+                if len(indexed_feature) < 2:
+                    continue
+
+                # Set feature value in row at the specified index
+                feature_index, feature_value = indexed_feature
+                feature_row[int(feature_index) - 1] = float(feature_value)
             feature_matrix.append(feature_row)
             target_vector.append(float(target))
         target_array = np.array(target_vector)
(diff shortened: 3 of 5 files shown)


.github/workflows/pip.yml
```diff
@@ -93,8 +93,11 @@ jobs:
           kh-samples sklearn -i khiops_classifier -e
 
           # Test that the line containing "MPI command" also contains
-          # "mpiexec", which means that `mpiexec` has been found
-          kh-status | grep "MPI command" | grep -wq mpiexec
+          # an executable name under a /bin directory
+          # Note: this executable name can be different, depending on the MPI
+          # backend and OS; for instance, "orterun" for OpenMPI on Ubuntu Linux, but
+          # "mpiexec" for OpenMPI on Rocky Linux
+          kh-status | grep "MPI command" | grep -Ewq "(/.+?)/bin/.+"
```
folmos-at-orange, line 100: This check is weaker than whaty is stated.

Maybe we should check `grep -Ewq "orterun|mpirun|mpiexec"` ?

.github/workflows/unit-tests.yml
```diff
@@ -177,6 +179,8 @@ jobs:
           # Force > 2 CPU cores to launch mpiexec
           KHIOPS_PROC_NUMBER: 4
         run: |-
+          # Make sure '/bin' is before '/usr/bin' in PATH
+          PATH=$(echo "/bin:"$PATH | sed 's#:/bin##')
```
folmos-at-orange, line 183: Empty line after this (paragraph)

tests/test_dataset_class.py
```diff
@@ -565,12 +565,13 @@ def _load_khiops_sparse_file(self, stream):
             target, features = line.split(b"\t")
             feature_row = np.zeros(100)
             for feature in features.strip().split(b" "):
-                feature_index, feature_value = feature.split(b":")
                 try:
-                    feature_value = float(feature_value)
-                # missing value, whence empty string
+                    feature_index, feature_value = feature.split(b":")
+                # missing value, whence there is no feature_value
                 except ValueError:
```
folmos-at-orange, line 571: I'd explain all this in a comment just before the `try`

.github/workflows/pip.yml
```diff
@@ -93,8 +93,11 @@ jobs:
           kh-samples sklearn -i khiops_classifier -e
 
           # Test that the line containing "MPI command" also contains
-          # "mpiexec", which means that `mpiexec` has been found
-          kh-status | grep "MPI command" | grep -wq mpiexec
+          # an executable name under a /bin directory
```
folmos-at-orange, line 96: General: Should this commit be squashed with the following ?

If not I'd rename it to `Add '/bin' before '/usr/bin' in PATH in CI`

.github/workflows/unit-tests.yml
```diff
@@ -99,6 +99,8 @@ jobs:
           OMPI_MCA_rmaps_base_oversubscribe: true
           PRTE_MCA_rmaps_default_mapping_policy: :oversubscribe
         run: |
+          # Make sure '/bin' is before '/usr/bin' in PATH
+          PATH=$(echo "/bin:"$PATH | sed 's#:/bin##')
```
folmos-at-orange, line 103: Empty line after this (paragraph)

folmos-at-orange (changes requested):

.github/workflows/pip.yml
popescu-v: Well, I am not sure: what if the executable is called something else (e.g. `mpiexec.hydra` for MPICH, whence we would also add this, or rather `mpiexec\.?.?` etc)?

It soon becomes brittle IMHO. In the current state, we just check that the MPI command points to something that is inside a `bin` directory. The tests that `mpiexec` is found are elsewhere (in the `unit-tests` workflow).

So, I would keep it this way; the comments contain "for instance" and this is key IMHO.

.github/workflows/unit-tests.yml
popescu-v: Ok.

tests/test_dataset_class.py
popescu-v: Ok, I will try to.

.github/workflows/pip.yml
popescu-v: I would keep the commits separate and change the commit message as suggested.

.github/workflows/unit-tests.yml
popescu-v: Ok.

.github/workflows/unit-tests.yml
popescu-v: Done.

tests/test_dataset_class.py
popescu-v: Done. Simplified and commented this code snippet.

.github/workflows/pip.yml
popescu-v: Done.

.github/workflows/unit-tests.yml
popescu-v: Done.

.github/workflows/pip.yml
folmos-at-orange: Ok. I think though the first part should be `(.+)`, and not `(.+?)`

popescu-v: Well, I wanted to accommodate the case when the path starts with `/bin` (this is not excluded in principle I guess).

folmos-at-orange: Ok

PR #226, opened 2024-08-07
191 test khiopspydev docker images before pushing them to the registry

### LINKED ISSUE #191: Improve Development Docker Image Build Workflow
#### Description
Currently, the `khiopspydev-*` Docker images are pushed to the GitHub container registry, each time with the `latest` tag, thus overriding previous Docker images. This hinders development of concomitant feature branches, where the updated Docker images (overridden for a specific feature branch) is accidentally used for launching the testing and packaging GitHub CI workflows. The goal of this issue is to streamline the Docker image building process, so that this dev chore disappears.

#### Questions/Ideas
- The `test` job can be removed: indeed, the Khiops binaries are now installed from upstream Khiops-generated sources (native OS-specific package, or Conda package) and are thus tested upstream.
- Versioning can be added to the Docker images, thus: `major.minor.patch.increment`, where `major`, `minor` and `patch` are from the `khiops` GitHub repository, and `increment` is zero-versioning-based, but orthogonal to the versioning scheme of `khiops-python` itself;
- Docker images tagged with specific versions can thus be pushed;
- Docker images are only pushed on manual `workflow_dispatch` events, when explicitly instructed to do so;
- Docker images are also tagged `latest` (and their tags pushed thereof) only if the workflow is manually launched from the `dev` or `main` branch (as these images are not "feature-branch" images anymore). Tag overriding is thus possible for the `latest` tag.
- The `unit-tests`, `pip` and `conda` workflows (which use the development Docker images) use the `latest` tag by default for these images, but allow for explicit overriding by manual launch via the `workflow_dispatch` events.

popescu-v:
closes #191

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -8,6 +8,9 @@ on:
         required: true
         type: boolean
         default: false
+      image-tag:
+        default: latest
+        description: Development Docker Image Tag
   pull_request:
     paths:
       - doc/**.rst
@@ -32,7 +35,11 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     container:
-      image: ghcr.io/khiopsml/khiops-python/khiopspydev-ubuntu22.04:latest
+      # 'latest' default image tag cannot be set as an environment variable,
+      # because the `env` context is only accessible at the step level;
+      # hence, it is hard-coded
+      image: |-
+        ghcr.io/khiopsml/khiops-python/khiopspydev-ubuntu22.04:${{ inputs.image-tag || 'latest' }}
       # Use the 'runner' user (1001) from github so checkout actions work properly
       # https://github.com/actions/runner/issues/2033#issuecomment-1598547465
       options: --user 1001
--- a/.github/workflows/dev-docker.yml
+++ b/.github/workflows/dev-docker.yml
@@ -2,6 +2,7 @@
 name: Dev Docker
 env:
   DEFAULT_KHIOPS_REVISION: 10.2.2
+  DEFAULT_IMAGE_INCREMENT: 0
   DEFAULT_SERVER_REVISION: main
   DEFAULT_PYTHON_VERSIONS: 3.8 3.9 3.10 3.11 3.12
 on:
@@ -13,13 +14,16 @@ on:
         type: string
         default: main
         description: Khiops Revision
+      image-increment:
+        type: string
+        description: Docker Image Tag Increment
       server-revision:
         type: string
         default: main
         description: Khiops Server Revision
       push:
         type: boolean
-        default: true
+        default: false
         description: Push to GH Registry
       python-versions:
         type: string
@@ -43,8 +47,10 @@ jobs:
         run: |
           set -x
           echo "KHIOPS_REVISION=${{ inputs.khiops-revision || env.DEFAULT_KHIOPS_REVISION }}" >> "$GITHUB_ENV"
+          echo "IMAGE_INCREMENT=${{ inputs.image-increment || env.DEFAULT_IMAGE_INCREMENT }}" >> "$GITHUB_ENV"
           echo "KHIOPSDEV_OS_CODENAME=$(echo '${{ matrix.khiopsdev-os }}' | tr -d '0-9.')" >> "$GITHUB_ENV"
           echo "SERVER_REVISION=${{ inputs.server-revision || env.DEFAULT_SERVER_REVISION }}" >> "$GITHUB_ENV"
+          echo "IMAGE_URL=ghcr.io/khiopsml/khiops-python/khiopspydev-${{ matrix.khiopsdev-os }}" >> "$GITHUB_ENV"
       - name: Checkout khiops-python sources
         uses: actions/checkout@v4
       - name: Set up Docker Buildx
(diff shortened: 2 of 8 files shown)


popescu-v:
> If I understood well, the docker image for dev will be still named (and tagged) this way :
> 
> ```
> khiopspydev-<OS among ubuntu22.04 and rocky9>:<khiops revision>.<image revision, number starting from 0>
> 
> If the image is built against the dev or main git branches an additional "latest" tag is set
> ```
> 
> It seems that the image revision is set manuallly by the user. There may be a risk that 2 users set the same image revision and thus overwrite the image. But it may be an acceptable risk.

Yes, IMHO it is an acceptable risk. Otherwise, we would need a different mechanism for ensuring no override occurs, e.g. use the commit hash in / as the image revision part of the Docker image tag (less readable)?

.github/workflows/api-docs.yml
```diff
@@ -1,5 +1,7 @@
 ---
 name: API Docs
+env:
+  DEFAULT_IMAGE_REVISION: latest
```
folmos-at-orange, line 4: I'd uniformize the use of either `_TAG`(`-tag`) or '_REVISION' (`-revision`) in all workflows/jobs. I'm inclined to 'TAG' since it is shorter.

.github/workflows/dev-docker.yml
```diff
@@ -2,6 +2,7 @@
 name: Dev Docker
 env:
   DEFAULT_KHIOPS_REVISION: 10.2.2
+  DEFAULT_IMAGE_REVISION: 0
```
folmos-at-orange, line 5: If the value is mandatory then we do not need a default value.

.github/workflows/api-docs.yml
```diff
@@ -29,10 +34,20 @@ concurrency:
...
   cancel-in-progress: false
 jobs:
+  set-image-tag:
+    runs-on: ubuntu-22.04
+    outputs:
+      image-tag: ${{ inputs.image-revision || env.DEFAULT_IMAGE_REVISION }}
+    steps:
+      - name: Set Docker image tag
+        run: |
+          echo "Setting development Docker image tag"
```
folmos-at-orange, line 44: This step is not necessary and it makes more complex the workflow. I'd just use
``${{ inputs.image-revision || env.DEFAULT_IMAGE_REVISION }}``.

.github/workflows/dev-docker.yml
```diff
@@ -56,6 +63,15 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Docker image tags
```
folmos-at-orange, line 66: `the Docker`

.github/workflows/pip.yml
```diff
@@ -17,6 +21,8 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-22.04
+    outputs:
+      image-tag: ${{ inputs.image-revision || env.DEFAULT_IMAGE_REVISION }}
```
folmos-at-orange, line 25: I'd just use the "or" incantation below instead of creating a job output. And in all jobs where you use this pattern.

folmos-at-orange (changes requested):

.github/workflows/api-docs.yml
popescu-v: I tried to, but it seems that `env` is not accessible here.

.github/workflows/pip.yml
popescu-v: At the beginning, I had tried to, but it seems that `env` is not accessible there.

.github/workflows/dev-docker.yml
popescu-v: Ok.

.github/workflows/api-docs.yml
popescu-v: Yes; we should however pay attention to the fact that sometimes we talk about the last field, some other times we talk about the whole tag x.y.z.REVISION. So, I would use TAG for the full tag and REVISION for its last field. But I would re-check all usages to make sure we're consistent here.

.github/workflows/dev-docker.yml
folmos-at-orange: After offline discussion this value is ok but it will be renamedto `DEFAULT_IMAGE_INCREMENT`

.github/workflows/api-docs.yml
popescu-v: Done: `INCREMENT` in `dev-docker` for the last part of the tag; `TAG` in the other workflows for the whole image tag.

.github/workflows/dev-docker.yml
popescu-v: Done. Kept the default value and removed the mandatory requirement.

.github/workflows/api-docs.yml
popescu-v: Done everywhere needed.

.github/workflows/dev-docker.yml
popescu-v: Done.

.github/workflows/pip.yml
popescu-v: This is now moot, as the `env` is not used anymore in this context (see above).

folmos-at-orange (approved):

folmos-at-orange (approved):

Issue #208, opened 2024-07-09
detect_format is not ignored when headerline and separator are provided


sgouache:
#### Description
In many API functions, the detect_format flag is available. It defaults to true, so users providing headerline and separator are sometimes unaware that is is set.
**Problem**: when both detect_format and headerline + separator are set, the detection is run even if its results are ignored. For some datatables, the detection produces confusing warnings which causes users to wonder if the processing was done with the right separator.
**Example output**:
warning : "/usr/local/lib/python3.10/dist-packages/runner/khiops_runner.py:465: UserWarning: Khiops ended correctly but there were minor issues: Warnings in log:
Line 2: warning : Database format detector s3://bkt-guporionis-backend-dev-ow-idh-fe/orionis/deployment/caller/2024/06/06/0800-1200/flows/*.csv : 1 other field separator is possible: '-'
  runner.run(args)".

#### Questions/Ideas
- Looking at the code it seems that outputs from the detection are ignored but detection is run nevertheless - so detect_format flag is not ignored 
- If this behavior is desired, the doc should be updated and user instructed to ignore such messages
- Else (preferred) fix the behavior of the flag: if both headerline and separator are set, detect format should be forced to false
- Also, the doc is misleading about the mixed case: if detect_format is true and only headerline is set, it will detect both and headerline set by the user will override the detection, allowing to perform partial detection

folmos-at-orange:
I'm looking into it. @sgouache could you please run the task with `trace=True` (or `--verbose` in `khrun`)

tramora:
A few general remarks (based on the latest 10.2.2.4 khiops-python release) regarding the so called **format spec preprocessing**...

Currently when you call **the python api** here are the results (they are consistent with the current docstring documentation) :

| Input | Output to khiops | Notes |
|---|---|---|
|*no input*|defect_format=True, header_line=True, field_separator=''|Room for improvement here (for human readers) : set header_line to False as detect_format will take precedence|
| header_line=True|defect_format=False, header_line=True, field_separator=''|
|field_separator=','|defect_format=False, header_line=True, field_separator=','|Room for improvement here, header_line should be left False as it does not correlate with the modified field|
| detect_format=True|defect_format=True, header_line=False, field_separator=''|
| detect_format=True, field_separator=',' |defect_format=False, header_line=False, field_separator=','|**cf python api vs protobuf api discussion below**|

[...]

PR #230, opened 2024-08-19
While preprocessing the datatable format specification remove any ambiguity

tramora:
Should (partially) address the issue #208. The remaining fix must be done in `khiops-api-base`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,8 @@ repos:
     hooks:
       - id: yamlfix
         exclude: packaging/conda/meta.yaml
+        language: python
+        additional_dependencies: [maison<2.0.0]
   - repo: https://github.com/python-jsonschema/check-jsonschema
     rev: 0.29.0
     hooks:
--- a/khiops/core/api.py
+++ b/khiops/core/api.py
@@ -237,13 +237,15 @@ def _preprocess_task_arguments(task_args):
         del task_args["use_complement_as_test"]
 
     # Preprocess the database format parameters
-    if "detect_format" in task_args:
-        assert "header_line" in task_args
-        assert "field_separator" in task_args
+    if (
+        task_args.get("detect_format")
+        or task_args.get("header_line")
+        or task_args.get("field_separator")
+    ):
         detect_format, header_line, field_separator = _preprocess_format_spec(
-            task_args["detect_format"],
-            task_args["header_line"],
-            task_args["field_separator"],
+            task_args.get("detect_format"),
+            task_args.get("header_line"),
+            task_args.get("field_separator"),
         )
         task_args["detect_format"] = detect_format
         task_args["header_line"] = header_line
@@ -317,15 +319,17 @@ def _preprocess_format_spec(detect_format, header_line, field_separator):
           their default values
         - It transforms the field separator "\t" to the empty string ""
     """
-    # Ignore detect_format if header_line or field_separator are set
-    if header_line is not None or field_separator is not None:
-        detect_format = False
 
     # Set the default values of header_line and field_separator
     if header_line is None:
-        header_line = True
+        header_line = False
     if field_separator is None:
         field_separator = ""
+
+    # Precedence rules :
+    # Ignore detect_format if header_line or field_separator is set to a non-empty value
+    if header_line or field_separator:
+        detect_format = False
 
     # Fail on separators with more than one char
     if len(field_separator) > 1:
@@ -510,12 +514,12 @@ def build_dictionary_from_data_table(
         Path of the output dictionary file.
     detect_format : bool, default ``True``
         If ``True`` detects automatically whether the data table file has a header and
-        its field separator. It's ignored if ``header_line`` or ``field_separator`` are
-        set.
-    header_line : bool, optional (default ``True`` if ``detect_format`` is ``False``)
+        its field separator.
+        It's set to ``False`` if ``header_line`` or ``field_separator`` are set.
+    header_line : bool, optional (default ``False``)
         If ``True`` it uses the first line of the data as column names. Overrides
         ``detect_format`` if set.
-    field_separator : str, optional (default "\\t" if ``detect_format`` is ``False``)
+    field_separator : str, optional (default ``""``)
         A field separator character, overrides ``detect_format`` if set ("" counts
         as "\\t").
     ... :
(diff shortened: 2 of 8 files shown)


khiops/core/api.py
```diff
@@ -317,16 +317,18 @@ def _preprocess_format_spec(detect_format, header_line, field_separator):
...
     # Set the default values of header_line and field_separator
     if header_line is None:
-        header_line = True
+        header_line = False
     if field_separator is None:
         field_separator = ""
 
+    # Precedence rules :
+    # Ignore detect_format if header_line or field_separator is set to a non-empty value
+    if header_line or (field_separator is not None and field_separator != ""):
```
popescu-v, line 329: IMHO, the condition can be simplified to `if header_line or field_separator`. Indeed, if `field_separator` is truthy, it means that it cannot be `None` or the empty string; conversely, if `field_separator` is `None` or the empty string, then it would be falsy.

khiops/core/api.py
```diff
@@ -317,16 +317,18 @@ def _preprocess_format_spec(detect_format, header_line, field_separator):
...
         - It transforms the field separator "\t" to the empty string ""
     """
-    # Ignore detect_format if header_line or field_separator are set
-    if header_line is not None or field_separator is not None:
-        detect_format = False
 
     # Set the default values of header_line and field_separator
     if header_line is None:
-        header_line = True
+        header_line = False
```
popescu-v, line 323: IMHO, the documentation should also be updated accordingly, if this is the desired behavior: in https://khiopsml.github.io/khiops-python/core/generated/khiops.core.api.html#module-khiops.core.api, the behavior of this flag can be clarified IMHO.

khiops/core/api.py
```diff
@@ -317,16 +317,18 @@ def _preprocess_format_spec(detect_format, header_line, field_separator):
...
     if field_separator is None:
         field_separator = ""
 
+    # Precedence rules :
+    # Ignore detect_format if header_line or field_separator is set to a non-empty value
+    if header_line or (field_separator is not None and field_separator != ""):
+        detect_format = False
```
popescu-v, line 330: IMHO, the documentation should be updated accordingly in https://khiopsml.github.io/khiops-python/core/generated/khiops.core.api.html#module-khiops.core.api : replace "it's ignored" with "it's set to `False`".

tests/test_core.py
```diff
@@ -772,22 +772,34 @@ def test_std_stream_errors(self):
 
 
 class MockedRunnerContext:
-    """A context to mock the `~.KhiopsLocalRunner.raw_run` function"""
+    """
+    A context to mock the low level `~.KhiopsLocalRunner.raw_run` function (by default)
```
popescu-v, line 776: I would try to use the `ScenarioWriteRunner` instead of mocking the high-level `run`, knowing that `task_args` are used for writing the scenarios: in that runner class, a specialized implementation of "high-level" `run` method exists already. I would try to update it, and the associated tests, for that matter.

popescu-v (changes requested):
See comments.

khiops/core/api.py
tramora: Simplified

khiops/core/api.py
tramora: Documentation updated

tramora:
I just noticed that @folmos-at-orange has already worked on this subject and created the [related branch](https://github.com/KhiopsML/khiops-python/tree/208-detect_format-is-not-ignored-when-headerline-and-separator-are-provided).
The 2 drafts are surely conflicting

tests/test_core.py
tramora: I seemed better and simplier to me to check the tasks args values before writing the scenario but you are right using the `ScenarioWriterRunner` will test all the process. It's cleaner afterwards and uses an existing runner for tests.
I rolled back the changes in `MockedRunnerContext` and adapted the test.

.pre-commit-config.yaml
```diff
@@ -28,6 +28,8 @@ repos:
     hooks:
       - id: yamlfix
         exclude: packaging/conda/meta.yaml
+        language: python
+        additional_dependencies: [maison<2.0.0]
```
popescu-v, line 32: IMHO, this should be in a separate commit (it is extraneous to the scope of the current PR, although needed).

khiops/core/api.py
```diff
@@ -237,17 +237,14 @@ def _preprocess_task_arguments(task_args):
...
-        detect_format, header_line, field_separator = _preprocess_format_spec(
-            task_args["detect_format"],
-            task_args["header_line"],
-            task_args["field_separator"],
-        )
-        task_args["detect_format"] = detect_format
-        task_args["header_line"] = header_line
-        task_args["field_separator"] = field_separator
+    detect_format, header_line, field_separator = _preprocess_format_spec(
+        task_args.get("detect_format", True),
```
popescu-v, line 241: Why do we need to enforce the `True` by default here? The default `True` value of the `detect_format` kwarg of the API functions is not enough?

tests/test_core.py
```diff
@@ -2818,6 +2818,72 @@ def test_khiops_mpi_command_args_var(self):
...
+            test_runner.subtest_name = test[0]
+            merged_kwargs = {
+                **{
+                    "additional_data_tables": {
+                        "Accident`Vehicles": "/tmp/Vehicules.txt"
+                    },
+                    "max_trees": 0,
+                },
+                **test[1],
+            }
```
popescu-v, line 2866: OK for now. Once Python 3.8 is dropped, we should be able to simplify this snippet to:

```python
merged_kwargs = {
    "additional_data_tables": {
        "Accident`Vehicles": "/tmp/Vehicules.txt"
    },
    "max_trees": 0,
} | test[1]
```
Perhaps adding a `TODO` comment in this respect would be useful?

.pre-commit-config.yaml
tramora: You are right, I splitted it in a different commit

khiops/core/api.py
tramora: It was a preparation of the change in the `khiops-api-base` (call of the `kh._preprocess_tasks_arguments` function) where I was not sure a default value of `detect_format` was set.

By the way I also fixed a problem for tasks that do not have format_spec in their signature.

tests/test_core.py
```diff
@@ -2818,6 +2818,72 @@ def test_khiops_mpi_command_args_var(self):
...
+            # 'detect_format': False, 'header_line': True, 'field_separator': ''
+            ("header_line", {"header_line": True}),
+            # 'field_separator': ',' ->
+            # 'detect_format': False, 'header_line': False, 'field_separator': ','
+            ("field_separator", {"field_separator": ","}),
+            # defect_format: True ->
+            # 'header_line': False, 'field_separator': '', defect_format: True
+            ("detect_format", {"detect_format": True}),
+            # defect_format: True, 'field_separator': ','
+            #     -> 'header_line': False, 'field_separator': '', defect_format: False
```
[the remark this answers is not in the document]
tramora: Fixed. Hopefully it was only in the comments

tests/test_core.py
tramora: You are right. I added the comment.

tramora:
To be closed as duplicated PR #234

PR #234, opened 2024-08-28
Factor out Khiops and task argument preprocessing in a separate function

### LINKED ISSUE #208: detect_format is not ignored when headerline and separator are provided

popescu-v:
closes #208

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -9,7 +9,7 @@ on:
         type: boolean
         default: false
       image-tag:
-        default: latest
+        default: 10.2.3-b.5.0
         description: Development Docker Image Tag
   pull_request:
     paths:
@@ -39,7 +39,7 @@ jobs:
       # because the `env` context is only accessible at the step level;
       # hence, it is hard-coded
       image: |-
-        ghcr.io/khiopsml/khiops-python/khiopspydev-ubuntu22.04:${{ inputs.image-tag || '10.2.3-b.4.with_conda-forge_10.2.3-b.4' }}
+        ghcr.io/khiopsml/khiops-python/khiopspydev-ubuntu22.04:${{ inputs.image-tag || '10.2.3-b.5.0' }}
       # Use the 'runner' user (1001) from github so checkout actions work properly
       # https://github.com/actions/runner/issues/2033#issuecomment-1598547465
       options: --user 1001
--- a/khiops/core/api.py
+++ b/khiops/core/api.py
@@ -122,27 +122,7 @@ def _run_task(task_name, task_args):
     stdout_file_path = task_args["stdout_file_path"]
     stderr_file_path = task_args["stderr_file_path"]
 
-    # Execute the preprocess of common task arguments
-    task_called_with_domain = _preprocess_task_arguments(task_args)
-
-    # Create a command line options object
-    command_line_options = CommandLineOptions(
-        batch_mode=task_args["batch_mode"] if "batch_mode" in task_args else True,
-        log_file_path=(
-            task_args["log_file_path"] if "log_file_path" in task_args else ""
-        ),
-        output_scenario_path=(
-            task_args["output_scenario_path"]
-            if "output_scenario_path" in task_args
-            else ""
-        ),
-        task_file_path=(
-            task_args["task_file_path"] if "task_file_path" in task_args else ""
-        ),
-    )
-
-    # Clean the task_args to leave only the task arguments
-    _clean_task_args(task_args)
+    command_line_options, task_called_with_domain = _preprocess_arguments(task_args)
 
     # Obtain the api function from the registry
     task = get_task_registry().get_task(task_name, get_khiops_version())
@@ -160,6 +140,43 @@ def _run_task(task_name, task_args):
     finally:
         if task_called_with_domain and not trace:
             fs.remove(task_args["dictionary_file_path"])
+
+
+def _preprocess_arguments(args):
+    """Preprocessing of Khiops arguments
+
+    Parameters
+    ----------
+    args : dict
+        The Khiops arguments.
+
+    Returns
+    -------
+    tuple
+        A 2-tuple containing:
+        - A `~.CommandLineOptions` instance
+        - A `bool` that is ``True`` if the value of the `dictionary_file_or_domain`
+          `args` key is a `~.DictionaryDomain` instance.
+
+    .. note:: This function *mutates* the input `args` dictionary.
+    """
+    # Execute the preprocess of common task arguments
+    task_is_called_with_domain = _preprocess_task_arguments(args)
+
+    # Create a command line options object
+    command_line_options = CommandLineOptions(
+        batch_mode=args["batch_mode"] if "batch_mode" in args else True,
+        log_file_path=(args["log_file_path"] if "log_file_path" in args else ""),
+        output_scenario_path=(
+            args["output_scenario_path"] if "output_scenario_path" in args else ""
+        ),
+        task_file_path=(args["task_file_path"] if "task_file_path" in args else ""),
+    )
+
+    # Clean the args to leave only the task arguments
+    _clean_task_args(args)
+
+    return command_line_options, task_is_called_with_domain
 
 
 def _preprocess_task_arguments(task_args):
--- a/khiops/core/api.py
+++ b/khiops/core/api.py
@@ -328,11 +328,11 @@ def _preprocess_format_spec(detect_format, header_line, field_separator):
     r"""Preprocess the user format spec to be used in a task
 
     More precisely:
-        - Disables ``detect_format`` if either ``header_line`` or ```field_separator``
-          are set
-        - If either ``header_line`` or ``field_separator`` is None then they are set to
-          their default values
-        - It transforms the field separator "\t" to the empty string ""
+        - Sets ``detect_format`` to ``False`` if either ``header_line`` or
+          ``field_separator`` are set
+        - If either ``header_line`` or ``field_separator`` is ``None``, then they are
+          set to their default values
+        - It transforms the field separator "\\t" to the empty string ""
     """
     # Ignore detect_format if header_line or field_separator are set
     if header_line is not None or field_separator is not None:
@@ -527,14 +527,16 @@ def build_dictionary_from_data_table(
         Path of the output dictionary file.
     detect_format : bool, default ``True``
         If ``True`` detects automatically whether the data table file has a header and
-        its field separator. It's ignored if ``header_line`` or ``field_separator`` are
-        set.
-    header_line : bool, optional (default ``True`` if ``detect_format`` is ``False``)
-        If ``True`` it uses the first line of the data as column names. Overrides
-        ``detect_format`` if set.
-    field_separator : str, optional (default "\\t" if ``detect_format`` is ``False``)
-        A field separator character, overrides ``detect_format`` if set ("" counts
-        as "\\t").
+        its field separator. It is set to ``False`` if ``header_line`` or
+        ``field_separator`` are set.
+    header_line : bool, optional (default ``True``)
+        If ``True`` it uses the first line of the data as column names. Sets
+        ``detect_format`` to ``False`` if set. Ignored if ``detect_format``
+        is ``True``.
+    field_separator : str, optional (default "\\t")
+        A field separator character. "" has the same effect as "\\t". Sets
+        ``detect_format`` to ``False`` if set. Ignored if ``detect_format``
+        is ``True``.
     ... :
         See :ref:`core-api-common-params`.
     """
@@ -580,14 +582,16 @@ def check_database(
         Path of the data table file.
     detect_format : bool, default ``True``
         If ``True`` detects automatically whether the data table file has a header and
-        its field separator. It's ignored if ``header_line`` or ``field_separator`` are
-        set.
-    header_line : bool, optional (default ``True`` if ``detect_format`` is ``False``)
-        If ``True`` it uses the first line of the data as column names. Overrides
-        ``detect_format`` if set.
-    field_separator : str, optional (default "\\t" if ``detect_format`` is ``False``)
-        A field separator character, overrides ``detect_format`` if set ("" counts
-        as "\\t").
+        its field separator. It is set to ``False`` if ``header_line`` or
+        ``field_separator`` are set.
+    header_line : bool, optional (default ``True``)
+        If ``True`` it uses the first line of the data as column names. Sets
+        ``detect_format`` to ``False`` if set. Ignored if ``detect_format``
+        is ``True``.
+    field_separator : str, optional (default "\\t")
+        A field separator character. "" has the same effect as "\\t". Sets
+        ``detect_format`` to ``False`` if set. Ignored if ``detect_format``
+        is ``True``.
     sample_percentage : float, default 100.0
         See the ``sampling_mode`` option below.
     sampling_mode : "Include sample" or "Exclude sample"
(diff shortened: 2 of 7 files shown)


khiops/core/api.py
```diff
@@ -107,6 +107,41 @@ def _get_or_create_execution_dictionary_file(dictionary_file_path_or_domain, tra
     return execution_dictionary_file_path
 
 
+def preprocess_arguments(args):
```
folmos-at-orange, line 110: For better locality move after `_run_task` and before `_preprocess_task_arguments`.

khiops/core/api.py
```diff
@@ -107,6 +107,41 @@ def _get_or_create_execution_dictionary_file(dictionary_file_path_or_domain, tra
...
+    """Preprocessing of Khiops arguments
+
+    Parameters
+    ----------
+    args : dict
+        The Khiops arguments.
+
+    Returns
+    -------
+    (`CommandLineOptions` instance, bool) tuple, with:
```
folmos-at-orange, line 120: The pattern in this type of documentation is 
```
tuple
    A 2-tuple containing:
    - A `~.CommandLineOptions` instance
    - A `bool` that is ``True`` whether the `dictionary_file_or_domain` is a `~.DictionaryDomain` instance
```

See for example `train_predictor`.

khiops/core/api.py
folmos-at-orange: Also, since this refactoring is only to acomodate `khrun` (which will be replaced) I'd mark the function as private (`_` prefix).

khiops/core/api.py
```diff
@@ -107,6 +107,41 @@ def _get_or_create_execution_dictionary_file(dictionary_file_path_or_domain, tra
...
+        output_scenario_path=(
+            args["output_scenario_path"] if "output_scenario_path" in args else ""
+        ),
+        task_file_path=(args["task_file_path"] if "task_file_path" in args else ""),
+    )
+
+    # Clean the args to leave only the task arguments
+    _clean_task_args(args)
+
+    return (command_line_options, task_is_called_with_domain)
```
folmos-at-orange, line 142: The parentheses are not necessary.

folmos-at-orange (changes requested):
See above.

khiops/core/api.py
popescu-v: Yes, right, my bad. I would fix this.

khiops/core/api.py
popescu-v: Indeed; I would remove them.

khiops/core/api.py
popescu-v: > For better locality move after `_run_task` and before `_preprocess_task_arguments`.

Ok.

popescu-v: > Also, since this refactoring is only to acomodate `khrun` (which will be replaced) I'd mark the function as private (`_` prefix). We will probably defactor this in the future.

Yeah, we could still use it while it is necessary, then de-factor it, indeed. However, having it public and properly documented would help with code understanding a(nd reuse as well), IMHO.

folmos-at-orange: This will be used only once in code that is not public. It is refactored only for convenience but is not part of the API (even the "internals" API).

khiops/core/api.py
popescu-v: Done.

khiops/core/api.py
popescu-v: Done for the first point. Not convinced for the second point (see my comment above).

khiops/core/api.py
popescu-v: Done.

khiops/core/api.py
popescu-v: Done & Done (both points).

folmos-at-orange (approved):

popescu-v:
Pushed alpha tag 10.2.2.5-a.1 on the feature branch of this PR.

tramora:
Thank you for the change.

I used this pre-release as a base for modifications in a POC version of `khiops-api-base/kh-run`

Indeed eventually it fixed the customer issue : https://github.com/KhiopsML/khiops/issues/345 because it now produces these directives 
```
TrainDatabase.HeaderLineUsed true
TrainDatabase.FieldSeparator
```
instead of the previous ones 
```
TrainDatabase.HeaderLineUsed true
TrainDatabase.FieldSeparator 
TrainDatabase.DatabaseFormatDetector.DetectFileFormat
```
So we do not face any longer the issue khiops core must fix in the next release.

A few remarks though we may have discussed already :

## Public function to use

Both `khiops.core.api._preprocess_arguments` and `khiops.core.api._preprocess_tasks_arguments` have the private _ prefix and thus by convention are not officially exposed. 

In `khiops-api-base/kh-run`  the second `khiops.core.api._preprocess_tasks_arguments` is to be used in a loop of the `api_calls`.

## Asserts & default values

Because of asserts in code, the caller has the burden to set the default values of the 2 parameters `header_line` and `field_separator` along with `detect_format` before calling the API. It would have been nice if the API could handle the "business rules" internally.

## Semantic

The initial input protobuf payload contains only the `detect_format=True` param. It is quite surprising to end with a `HeaderLineUsed true` in the generated scenario even if the docstring states it clearly (I was forced to set the default values of `header_line` and `field_separator`)

```
It's ignored if ``header_line`` or ``field_separator`` are set.
```

tramora:
As the produced output was incorrect (see the previous comment about the semantic), I tweaked in the last commit the `khiops-api-base/kh-run` code in order to both keep the semantic and to try to avoid ambiguity for khiops.

However the ambiguities remain (in **bold** below) and we are waiting for a khiops release to fix the issue https://github.com/KhiopsML/khiops/issues/345 about the `DetectFileFormat` 

| Input to `khiops-api-base/kh-run`| Output to khiops after processing by the API | Notes |
|---|---|---|
|*no input*|defect_format=True, header_line=True, field_separator=''|**Room for improvement here (for human readers) : set header_line to False as detect_format will take precedence**|
| header_line=True|defect_format=False, header_line=True, field_separator=''|
|field_separator=','|defect_format=False, header_line=True, field_separator=','|**Room for improvement here, header_line should be left False as it does not correlate with the modified field**|
| detect_format=True|defect_format=True, header_line=True, field_separator=''|**Room for improvement here (for human readers) : set header_line to False as detect_format will take precedence**|
| detect_format=True, field_separator=',' |defect_format=False, header_line=True, field_separator=','|**Room for improvement here, header_line should be left False as it does not correlate with the modified field**|

popescu-v:
> Thank you for the change.
> 
> I used this pre-release as a base for modifications in a POC version of `khiops-api-base/kh-run`
> 
> Indeed eventually it fixed the customer issue : [KhiopsML/khiops#345](https://github.com/KhiopsML/khiops/issues/345) because it now produces these directives
> 
> ```
> TrainDatabase.HeaderLineUsed true
> TrainDatabase.FieldSeparator
> ```
> 
> instead of the previous ones
> 
> ```
> TrainDatabase.HeaderLineUsed true
> TrainDatabase.FieldSeparator 
> TrainDatabase.DatabaseFormatDetector.DetectFileFormat
> ```
> 
> So we do not face any longer the issue khiops core must fix in the next release.
> 
> A few remarks though we may have discussed already :
> ## Public function to use
> 
> Both `khiops.core.api._preprocess_arguments` and `khiops.core.api._preprocess_task_arguments` have the private _ prefix and thus by convention are not officially exposed.
> 
> In `khiops-api-base/kh-run` the second `khiops.core.api._preprocess_task_arguments` is to be used in a loop of the `api_calls`.
> 
> Branch : https://gitlab.tech.orange/khiops/khiops-api-base/-/tree/preprocess_task_arguments?ref_type=heads
> ## Asserts & default values
> 
> Because of asserts in code, the caller has the burden to set the default values of the 2 parameters `header_line` and `field_separator` along with `detect_format` before calling the API. It would have been nice if the API could handle the "business rules" internally.
> ## Semantic
> 
> The initial input protobuf payload contains only the `detect_format=True` param. It is quite surprising to end with a `HeaderLineUsed true` in the generated scenario even if the docstring states it clearly (I was forced to set the default values of `header_line` and `field_separator`).
> 
> ```
> (detect_format) ignored if ``header_line`` or ``field_separator`` are set.
> ```

Yes, here it should be ```is set to ``False`` if `header_line`` or ``field_separator`` are set```. On the other hand, if `header_line None`, then `header_line` is set to `True` in all cases; this where the documentation seems inexact, as it states that this should be the case only if `detect_format` is set to `False`.

khiops/core/api.py
```diff
@@ -328,11 +328,11 @@ def _preprocess_format_spec(detect_format, header_line, field_separator):
...
-        - Disables ``detect_format`` if either ``header_line`` or ```field_separator``
-          are set
-        - If either ``header_line`` or ``field_separator`` is None then they are set to
-          their default values
-        - It transforms the field separator "\t" to the empty string ""
+        - Sets ``detect_format`` to ``False`` if either ``header_line`` or
+          ``field_separator`` are set
+        - If either ``header_line`` or ``field_separator`` is ``None``, then they are
+          set to their default values
+        - It transforms the field separator "\\t" to the empty string ""
```
folmos-at-orange, line 335: The second `\` is not necessary: we are in a raw string `r"..."`

khiops/core/api.py
```diff
@@ -527,14 +527,15 @@ def build_dictionary_from_data_table(
...
-        as "\\t").
+        its field separator. It is set to ``False`` if ``header_line`` or
+        ``field_separator`` are set.
+    header_line : bool, optional (default ``None``)
+        It is set to ``True`` if ``None``.
+        If ``True`` it uses the first line of the data as column names. Sets
+        ``detect_format`` to ``False`` if set.
+    field_separator : str, optional (default ``None``)
+        A field separator character. It is set to "" if ``None``. "" has the same
+        effect as "\\t". Sets ``detect_format`` to ``False`` if set.
```
folmos-at-orange, line 538: I'd keep the old phrasing of the default value ("default `True` if ..."). This way the user is immediately informed about the default value and doesn't have to look below for the side effect.

I'm ok with the "Sets `detect_format` .. to ``False``" change.

khiops/core/api.py
folmos-at-orange, line 1: General comment on the doc commit: Why the message says that in _now_ reflects the current behavior? It is just a rephrasing.

khiops/core/api.py
folmos-at-orange: And all below.

.github/workflows/api-docs.yml
```diff
@@ -9,7 +9,7 @@ on:
         type: boolean
         default: false
       image-tag:
-        default: latest
+        default: 10.2.3-b.5.0
```
folmos-at-orange, line 12: General Comment: I'd rephrase the commit
Set _in CI_ Khiops core packages version to the latest beta

folmos-at-orange (changes requested):

khiops/core/api.py
popescu-v: IMHO, no: before it was stated that:
- `detect_format` is ignored if at least one of `header_line` or `field_separator` is set; whereas, in reality, it set to `False`, which has the consequence of not performing automatic format detection on the binary Khiops side.
- `header_line` has `True` as default value, which is not strictly true, because the default value is `None`, in which case the attribute is set to `True` *afterwards*; the current discourse seems misleading, in that it creates the expectation that the default value is `True`;
- similar situation for `field_separator`, where the default value is not "\t", but `None`, in which case the attribute is set to "" *afterwards*; and "\t" and "" are equivalent *for the Khiops binaries* only.

So, the goal of this commit is to make sure the documentation reflects current behavior and semantics.

.github/workflows/api-docs.yml
popescu-v: Yes, OK for the addition, thus the intent is clearer.

khiops/core/api.py
popescu-v: IMHO, it is better that the documentation reflects *exactly* the behavior (w/o shortcuts); otherwise, it seems misleading (see the other related comment). So, I would keep `None` as default value, with the statement of the side effect: longer-winded, but more correct IMHO. And IMHO this becomes even more important for client usage.

As for breaking the pattern for `header_line`: IMHO we are obliged to do so, in order to establish the "link" between the default `None` and the consequent default `True` as a side-effect. Otherwise, the documentation is not faithful to the implemented behavior IMHO.

khiops/core/api.py
popescu-v: Ok, I will recheck in all cases and change where mandated by the `r"..."`.

popescu-v: Done.

.github/workflows/api-docs.yml
popescu-v: Done.

khiops/core/api.py
popescu-v: Done as discussed.

khiops/core/api.py
popescu-v: Removed the "reflects the current behavior" part.

khiops/core/api.py
```diff
@@ -510,14 +527,16 @@ def build_dictionary_from_data_table(
...
-    field_separator : str, optional (default "\\t" if ``detect_format`` is ``False``)
-        A field separator character, overrides ``detect_format`` if set ("" counts
-        as "\\t").
+        its field separator. It is set to ``False`` if ``header_line`` or
+        ``field_separator`` are set.
+    header_line : bool, optional (default ``True``)
+        If ``True`` it uses the first line of the data as column names. Sets
+        ``detect_format`` to ``False`` if set. Ignored if ``detect_format``
+        is set to ``True``.
+    field_separator : str, optional (default "")
```
folmos-at-orange, line 536: Set the default value to `\t` as before. Again, this _is_ its semantic value. And all below.

khiops/core/api.py
```diff
@@ -510,14 +527,16 @@ def build_dictionary_from_data_table(
...
+        A field separator character. "" has the same effect as "\t". Sets
+        ``detect_format`` to ``False`` if set. Ignored if ``detect_format``
+        is set to ``True``.
```
folmos-at-orange, line 539: -> ```Ignored if ``detect_format`` is ``True`` ```.

tramora:
Go for it for this PR despite the ambiguities remaining : https://github.com/KhiopsML/khiops-python/pull/234#issuecomment-2411385695

However the issue https://github.com/KhiopsML/khiops/issues/345 on the khiops core side is not fixed yet (in `10.2.3-b.5`) so the final state remains the same.

tramora (approved):

folmos-at-orange (changes requested):

khiops/core/api.py
popescu-v: Ok & Done.

khiops/core/api.py
popescu-v: Ok & Done.

PR #237, opened 2024-09-02
Eliminate "v" prefix from samples version

folmos-at-orange:
See https://github.com/KhiopsML/khiops/issues/234

### DIFF
--- a/khiops/tools.py
+++ b/khiops/tools.py
@@ -88,6 +88,10 @@ def kh_samples_entry_point():  # pragma: no cover
         sys.exit(1)
 
 
+# Samples version: To be updated when khiops-samples does
+DEFAULT_SAMPLES_VERSION = "10.2.0"
+
+
 def kh_download_datasets_entry_point():
     """Entry point for the download samples helper script"""
     samples_dir = pathlib.Path.home() / "khiops_data" / "samples"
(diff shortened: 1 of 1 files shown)


khiops/tools.py
```diff
@@ -96,7 +96,7 @@ def kh_download_datasets_entry_point():
         description=f"Downloads the Khiops samples dataset to {samples_dir}.",
     )
     arg_parser.add_argument(
-        "-v", "--version", default="10.1.1", help="Sample datasets version"
+        "-v", "--version", default="10.2.0", help="Sample datasets version"
```
popescu-v, line 99: I am not sure about keeping this default here, as we would always need to bump it to the latest release.

folmos-at-orange: Maybe a global variable ? We need both default values because the tool may be called from a shell or from Python.

popescu-v: Yes, I see your point. However, it seems problematic to me to hard-code this version, thusly needing to update this code each time we bump the release version of the samples.

What if we do not add a default version and just make this argument required, thus leaving on the consumers of the tool the responsibility of setting this correctly?

popescu-v (commented):
See comment.

khiops/tools.py
folmos-at-orange: This is intended to be easy to use, so no to add an argument.

popescu-v: OK, then this global variable would be used for setting both:
- `version` default value in `download_datasets`; and
- `-v` argument in the `kh_download_datasets_entry_point`.
Hence, I would close this thread as soon as this change is implemented.

popescu-v (approved):
LGTM

PR #247, opened 2024-09-18
Change reference path from khiops.__file__ to __file__

folmos-at-orange:
We used this value to obtain the package installation path. However if the program is execute in a directory containing a `khiops` directory it is set to other values (usually `None`).

The now we use only `__file__` in `runner.py` which ensures that we have a true reference point to find the package path.


popescu-v (approved):
LGTM

PR #249, opened 2024-09-18
Add set latest docker dev

folmos-at-orange:


.github/workflows/dev-docker.yml
```diff
@@ -16,19 +16,24 @@ on:
         description: Khiops Revision
       image-increment:
         type: string
-        description: Docker Image Tag Increment
-      server-revision:
-        type: string
-        default: main
-        description: Khiops Server Revision
+        default: '0'
```
popescu-v, line 19: I don't think we need quotes here (see above).

.github/workflows/dev-docker.yml
```diff
@@ -63,15 +67,16 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Set up the Docker Image Tags
-        run: |-
+        run: |
           set -x
-          if [[ "$GITHUB_REF" == "refs/heads/main" || "$GITHUB_REF" == "refs/heads/dev" ]]
```
popescu-v, line 68: IMHO we want to be able to set the `latest` tag on request, *only when* on `main` or `dev` (i.e. when the feature branch has been merge). I don't think we want to be able to set `latest` on a feature branch, because `latest` being the default tag, other feature branches might still accidentally use a per-feature branch `latest`. So, I would reinstate the previous code and add the new condition as `&&`.

popescu-v (changes requested):
See comments (among which one important one, IMHO).

.github/workflows/dev-docker.yml
```diff
@@ -2,7 +2,7 @@
 name: Dev Docker
 env:
   DEFAULT_KHIOPS_REVISION: 10.2.2
-  DEFAULT_IMAGE_INCREMENT: 0
+  DEFAULT_IMAGE_INCREMENT: '0'
```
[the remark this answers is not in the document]
folmos-at-orange: Ok, changed to number.

.github/workflows/dev-docker.yml
folmos-at-orange: Ok as above.

.github/workflows/dev-docker.yml
folmos-at-orange: Ok & done.

popescu-v: Ok, right.

.github/workflows/dev-docker.yml
popescu-v: Ok, good.

popescu-v (approved):
LGTM. Can be merged IMHO, via a merge commit.

PR #250, opened 2024-09-26
When openmpi is launched with the quiet flag, the return code must be handled though

tramora:
In version 10.2.2.4, openmpi is launched with the `--quiet` flag that shallows every error message.

It seems safe to remove this flag to give informative messages to the end user.
If the flag is to be kept the return code must be evaluated otherwise openmpi exits silently without letting the user know the issue (for example when the proc number is too high)

### DIFF
--- a/khiops/core/internals/runner.py
+++ b/khiops/core/internals/runner.py
@@ -891,6 +891,13 @@ class KhiopsRunner(ABC):
                     f"Khiops ended correctly but there were minor issues:\n{error_msg}"
                 )
                 warnings.warn(error_msg.rstrip())
+        else:
+            # no message available, an error must be raised though
+            if return_code != 0:
+                raise KhiopsRuntimeError(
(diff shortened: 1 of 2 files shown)


popescu-v:
IMHO, the way to go is to:
1/ Wait for https://github.com/KhiopsML/khiops-python/issues/175 to land into `khiops` (after release 10.2.3).
2/ Document the proper environment variable configuration to set, in order to tune the (Open)MPI verbosity.
Hence, there is no need to modify `khiops-python` in this respect IMHO.

Pending this change, one temporary "escape hatch" could be to update the `KHIOPS_MPI_COMMAND` environment variable via a command like:
`export KHIOPS_MPI_COMMAND=$(echo $KHIOPS_MPI_COMMAND | sed 's/ --quiet//')`
Thusly, the `--quiet` flag would be locally removed.

If this solution is OK, then this PR could be closed IMHO.

PR #255, opened 2024-10-09
253 Raise a visible error in the border case where no error message is available

### LINKED ISSUE #253: Should raise a visible error in the border case where no error message is available
#### Description
In a very few border cases (kill of a underlying process at the launch time, `openmpi` facing an error but launched with the --quiet option, etc...) the current 10.2.2.4 version the API can end silently without raising any error.

The end user cannot detect then that an error occurred.

#### Questions/Ideas
These cases should be detected and reported via a KhiopsRuntimeError exception even if no error message can be provided.

tramora:
Should fix #253

### DIFF
--- a/khiops/core/internals/runner.py
+++ b/khiops/core/internals/runner.py
@@ -746,6 +746,13 @@ class KhiopsRunner(ABC):
                     f"Khiops ended correctly but there were minor issues:\n{error_msg}"
                 )
                 warnings.warn(error_msg.rstrip())
+        # Raise an exception anyway for a non-zero return code without any message
+        else:
+            if return_code != 0:
+                raise KhiopsRuntimeError(
+                    f"{tool_name} execution had errors (return code {return_code}) "
+                    "but no message is available"
+                )
 
     def _collect_errors(self, log_file_path):
         # Collect errors any errors found in the log
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -18,6 +18,7 @@ from unittest import mock
 
 import khiops
 import khiops.core as kh
+from khiops.core import KhiopsRuntimeError
 from khiops.core.internals.common import create_unambiguous_khiops_path
 from khiops.core.internals.io import KhiopsOutputWriter
 from khiops.core.internals.runner import KhiopsLocalRunner, KhiopsRunner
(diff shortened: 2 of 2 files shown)


khiops/core/internals/runner.py
```diff
@@ -746,6 +746,13 @@ def _report_exit_status(
                     f"Khiops ended correctly but there were minor issues:\n{error_msg}"
                 )
                 warnings.warn(error_msg.rstrip())
+        else:
+            # no message available, an error must be raised though
+            if return_code != 0:
+                raise KhiopsRuntimeError(
+                    f"{tool_name} execution had errors (return code {return_code}) "
+                    "but no message is available\n"
```
folmos-at-orange, line 754: Remove the `\n`

khiops/core/internals/runner.py
```diff
@@ -746,6 +746,13 @@ def _report_exit_status(
                     f"Khiops ended correctly but there were minor issues:\n{error_msg}"
                 )
                 warnings.warn(error_msg.rstrip())
+        else:
+            # no message available, an error must be raised though
```
folmos-at-orange, line 750: I'd put `Raise an exception anyway if the return code was not zero with no messages`

tests/test_core.py
```diff
@@ -2691,6 +2692,24 @@ def test_khiops_environment_variables_basic(self):
                 else:
                     os.environ[fixture["variable"]] = old_value
 
+    def test_raise_exception_on_error_case_without_a_message(self):
+        with self.assertRaises(KhiopsRuntimeError):
+            with MockedRunnerContext(
+                create_mocked_raw_run(
+                    False,  # ask for an empty stdout
+                    False,  # ask for an empty stderr
+                    9,  # non-zero error code
```
folmos-at-orange, line 2701: You can use named parameters to self document:
```
create_mocked_raw_run(
    stdout=False,
    stderr=False,
    return_code=9
)

**Review by folmos-at-orange — APPROVED** [review]

File: khiops/core/internals/runner.py (hunk 1 above)
Comment (tramora) [reply khiops/core/internals/runner.py]: You are right. I removed it

File: khiops/core/internals/runner.py (hunk 2 above)
Comment (tramora) [reply khiops/core/internals/runner.py]: Indeed clearer. Modification done

File: tests/test_core.py (hunk 3 above)
Comment (tramora) [reply tests/test_core.py]: Indeed I used the named params but kept the comments

File: tests/test_core.py [hunk 4]
```diff
@@ -2691,6 +2692,24 @@ def test_khiops_environment_variables_basic(self):
                 else:
                     os.environ[fixture["variable"]] = old_value
 
+    def test_raise_exception_on_error_case_without_a_message(self):
+        with self.assertRaises(KhiopsRuntimeError):
```
popescu-v, line 2696: I would also test the error message:
```python
        with self.assertRaises(KhiopsRuntimeError) as context:
            # The rest of the code as-is
        expected_msg = "khiops execution had errors (return code 9) but no message is available"
        output_msg = str(context.exception )
        self.assertEqual(output_msg, expected_msg)
```

popescu-v (changes requested):
See comment.

khiops/core/internals/runner.py
folmos-at-orange: I'd put it before the `else` to follow the usual codebase pattern.

tests/test_core.py
tramora: Indeed it ensures the `KhiopsRuntimeError` was raised for the specific cause not because of another one.
However I'm reluctant to test against labels vs ids (because of i18n or encoding differences that can make the test fail) but it's OK here.
Change done.

popescu-v: Ok.

khiops/core/internals/runner.py
tramora: Modification done. Even if I would naturally keep it close to test against the return_code

popescu-v (approved):
LGTM

PR #262, opened 2024-10-16
Stop setting OpenMPI-specific environment variable

popescu-v:
Thus, the invariant that only `khiops_env` sets all required environment for Khiops is observed.

Any variable that is necessary should be added to `khiops_env`. However, following tests, the `OMPI_MCA_plm_rsh_agent` does not seem to be needed anymore.


folmos-at-orange (approved):
I'd rename the commit `Stop setting OpenMPI SSH related environment variable`

popescu-v:
> I'd rename the commit `Stop setting OpenMPI SSH related environment variable`

Ok & Done.

PR #271, opened 2024-10-30
Shorten split helper function import path

folmos-at-orange:


.github/workflows/api-docs.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         with:
           repository: khiopsml/khiops-python-tutorial
           path: doc/khiops-python-tutorial
+          ref: shorten-helpers-path
```
popescu-v, line 56: After approval of this PR, should merge `shorten-helpers-path` into `main` in `khiops-python-tutorial`, then merge the current PR, right?

folmos-at-orange: Yes, this is in a `TMP` commit.

tests/test_helper_functions.py
```diff
@@ -16,7 +16,7 @@
 
 from khiops.core.dictionary import DictionaryDomain
 from khiops.core.helpers import build_multi_table_dictionary_domain, visualize_report
-from khiops.utils.helpers import train_test_split_dataset
+from khiops.sklearn.helpers import train_test_split_dataset
```
popescu-v, line 19: IMHO this should be 
```
from khiops.sklearn import train_test_split_dataset
```,
because `train_test_split_dataset`  is imported in the `sklearn` package namespace.

File: doc/create-doc [hunk 3]
```diff
@@ -79,6 +79,7 @@ fi
 
 # Convert tutorials to reST
 tutorials_dir="./tutorials"
+tutorials_dir="$(realpath $tutorials_dir)"
```
popescu-v, line 82: Why not conflate these two lines into a single one, rather than overriding `tutorials_dir`?

tests/test_helper_functions.py
folmos-at-orange: It tests this is less important, but ok.

doc/create-doc
folmos-at-orange: Ok

tests/test_helper_functions.py
popescu-v: > It tests this is less important, but ok.

IMHO, it allows to validate the coherence between samples (not always run) and tests (always run).

folmos-at-orange: Done

doc/create-doc
folmos-at-orange: Done

popescu-v (approved):
LGTM.
Before merge: remove `TMP` commit.
After merge: also merge `shorten-helpers-path` into `main` in `khiops-python-tutorial`.

PR #272, opened 2024-10-31
Fix FileTable to use only `filesystems` module

folmos-at-orange:
https://github.com/KhiopsML/khiops-python/pull/272#issuecomment-2450358485


folmos-at-orange:
Fix implemente to unblock #175 . `FileTable` is going to be eliminated by V11.

popescu-v (approved):
LGTM

PR #311, opened 2024-12-12
Update pairs documentation

folmos-at-orange:
_Put your message here_

---

### TODO Before Asking for a Review
- [ ] Rebase your branch to the latest version of `dev` (or `main` for release PRs)
- [ ] Make sure all CI workflows are green
- [ ] When adding a public feature/fix: Update the `Unreleased` section of `CHANGELOG.md` (no date)
- [ ] Self-Review: Review "Files Changed" tab and fix any problems you find
- API Docs (only if there are changes in docstrings, rst files or samples):
  - [ ] Check the docs build **without** warning: see the log of the API Docs workflow
  - [ ] Check that your changes render well in HTML: download the API Docs artifact and open `index.html`
  - If there are any problems it is faster to iterate by [building locally the API Docs](../blob/dev/doc/README.md#build-the-documentation)

PR #313, opened 2024-12-16
310 eliminate pair support from khiopsregressor

folmos-at-orange:
commit 3eb545cef17f538ac1c212650b996ca6ce8d38dc (HEAD -> 310-eliminate-pair-support-from-khiopsregressor, origin/310-eliminate-pair-support-from-khiopsregressor)
Author: Felipe Olmos <92923444+folmos-at-orange@users.noreply.github.com>
Date:   Thu Dec 12 15:57:24 2024 +0100

    Remove n_pairs parameter from KhiopsRegressor

    It was never supported.

commit 1ba74d2e82d3773afb3e407cee5f70cfd907b534
Author: Felipe Olmos <92923444+folmos-at-orange@users.noreply.github.com>
Date:   Thu Dec 12 15:56:51 2024 +0100

    Update pairs documentation

---

### DIFF
--- a/khiops/sklearn/estimators.py
+++ b/khiops/sklearn/estimators.py
@@ -198,6 +198,37 @@ def _check_numerical_target_type(ds):
         )
 
 
+def _check_pair_parameters(estimator):
+    assert isinstance(estimator, (KhiopsClassifier, KhiopsEncoder)), type_error_message(
+        "estimator", estimator, KhiopsClassifier, KhiopsEncoder
+    )
+    if not isinstance(estimator.n_pairs, int):
+        raise TypeError(type_error_message("n_pairs", estimator.n_pairs, int))
+    if estimator.n_pairs < 0:
+        raise ValueError("'n_pairs' must be positive")
+    if estimator.specific_pairs is not None:
+        if not is_list_like(estimator.specific_pairs):
+            raise TypeError(
+                type_error_message(
+                    "specific_pairs", estimator.specific_pairs, "list-like"
+                )
+            )
+        else:
+            for pair in estimator.specific_pairs:
+                if not isinstance(pair, tuple):
+                    raise TypeError(type_error_message(pair, pair, tuple))
+    if not isinstance(estimator.all_possible_pairs, bool):
+        raise TypeError(
+            type_error_message("all_possible_pairs", estimator.all_possible_pairs, bool)
+        )
+
+    # Check 'group_target_value' parameter
+    if not isinstance(estimator.group_target_value, bool):
+        raise TypeError(
+            type_error_message("group_target_value", estimator.group_target_value, bool)
+        )
+
+
 def _cleanup_dir(target_dir):
     """Cleanups a directory with only files in it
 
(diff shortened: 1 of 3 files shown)


popescu-v:
> Waiting for #311 to be merged.

I think it can be, as of now.

khiops/sklearn/estimators.py
```diff
@@ -2903,6 +2890,27 @@ def _fit_check_params(self, ds, **kwargs):
         # Call parent method
         super()._fit_check_params(ds, **kwargs)
 
+        # Check the pair related parameters
```
popescu-v, line 2893: These checks are duplicated between the Classifier and the Encoder. Couldn't we factor them out (in a mixin perhaps)?

popescu-v (changes requested):
Factor out pair-specific checks so that they can be mutualized between the KhiopsClassifier and the KhiopsEncoder.

khiops/sklearn/estimators.py
folmos-at-orange: Very difficult due to the class structure. It is simpler to duplicate the code.
```
    KhiopsEstimator(ABC, BaseEstimator)
        |
        +- KhiopsCoclustering(ClusterMixin)
        |
        +- KhiopsSupervisedEstimator
           |
           +- KhiopsPredictor
           |  |
           |  +- KhiopsClassifier(ClassifierMixin)
           |  |
           |  +- KhiopsRegressor(RegressorMixin)
           |
           +- KhiopsEncoder(TransformerMixin)
```

popescu-v: IMHO, we could use 2 mixins for `KhiopsEncoder` and `KhiopsClassifier`, the second mixin containing just this common code. It seems like adding complexity, I can realize that, but at the same time it would make the intent clearer and ease maintenance because it would factor out this common code in a single place, while using Pythonic reuse mechanisms.

The alternative to the second mixin would be to use "bare" composition by defining a standalone object which would take an estimator instance as parameter and contain the pair-specific logic, and would be created and used in each of the two estimator classes (it would take `self` as a parameter and switch on the type of `self`: if encoder or classifier, apply the logic; else, raise `NotImplementedError` or something along these lines). Or this could be a "private" class and the constraints on the accepted estimator parameter could be checked via `assert`s.

Either of these seems preferable to code duplication as it currently stands.

folmos-at-orange: Both of your solutions solve the code duplication at the expense of more layers of indirection. The code it is already _very_ indirect, due to the template pattern we use.

The proposed code is local, small and it is duplicated only two times. I can add comments to warn future devs refactoring this code. A little duplication is better than more layers of indirection.

The thing is, the hierarchy is a very tight abstraction for our classes. We could have a more flexible one by composition, but as discussed offline it is maybe too late.

popescu-v: Ok, I can see your point about the supplemental layer of indirection. Then, how about just one (or a couple of) function(s) that take `self` (the estimator instance) as parameter for doing the checks?

folmos-at-orange: Ok & Done

popescu-v (approved):
LGTM

PR #315, opened 2024-12-18
Handle `khiops_env --env` Execution Errors

### LINKED ISSUE #300: `khiops_env --env` Execution Error is Not (Properly) Handled
#### Description
Currently, `khiops_env --env` is executed with the appropriate path to the script, but only `stdout` is captured. Neither `stderr`, nor `returncode` are captured and handled.

Thus, if `khiops_env --env` fails for some reason (see e.g. https://github.com/KhiopsML/khiops/issues/444, most notably comment https://github.com/KhiopsML/khiops/issues/444#issuecomment-2505656344), there is no hint provided to the user. Just that the environment would not be set, which would entail failure later on.

#### Questions/Ideas
- The solution is to capture `stderr` and `.returncode` from the `khiops_env --env` subprocess in `_initialize_khiops_environment`, so that the relevant `KhiopsEnvironmentError` is raised, with the message contained in the process' `stderr`.

popescu-v:
closes #300 

---

### DIFF
--- a/khiops/core/internals/runner.py
+++ b/khiops/core/internals/runner.py
@@ -842,7 +842,12 @@ class KhiopsLocalRunner(KhiopsRunner):
             stderr=subprocess.PIPE,
             universal_newlines=True,
         ) as khiops_env_process:
-            stdout, _ = khiops_env_process.communicate()
+            stdout, stderr = khiops_env_process.communicate()
+            if khiops_env_process.returncode != 0:
+                raise KhiopsEnvironmentError(
+                    "Error initializing the environment for Khiops from the "
+                    f"{khiops_env_path} script. Contents of stderr:\n{stderr}"
+                )
             for line in stdout.split("\n"):
                 tokens = line.rstrip().split(maxsplit=1)
                 if len(tokens) == 2:
--- a/tests/test_khiops_integrations.py
+++ b/tests/test_khiops_integrations.py
@@ -9,7 +9,9 @@
 import os
 import platform
 import shutil
+import stat
 import subprocess
+import tempfile
 import unittest
 
 import khiops.core as kh
@@ -94,6 +96,76 @@ class KhiopsRunnerEnvironmentTests(unittest.TestCase):
             self.assertTrue(os.access(mpiexec_path, os.X_OK))
         else:
             self.skipTest("Skipping test: MPI support is not installed")
+
+    def test_environment_error_on_bogus_khiops_env_script(self):
+        """Test that error is raised on reading erroneous khiops_env script"""
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+
+            # Create temporary khiops_env binary dir
+            # Note: The "bin" subdir is needed for Windows.
+            temp_khiops_env_dir = os.path.join(temp_dir, "bin")
+            os.makedirs(temp_khiops_env_dir)
+
+            # Create temporary khiops_env path
+            temp_khiops_env_file_path = os.path.join(temp_khiops_env_dir, "khiops_env")
+
+            # On Windows, set KHIOPS_HOME to the temp dir
+            original_khiops_home_env_var = os.environ.get("KHIOPS_HOME")
+            if platform.system() == "Windows":
+                os.environ["KHIOPS_HOME"] = temp_dir
+                temp_khiops_env_file_path += ".cmd"
+
+            # Replace the khiops_env with a script that fails showing an error message
+            with open(
+                temp_khiops_env_file_path, "w", encoding="utf-8"
+            ) as temp_khiops_env_file:
+                test_error_message = "Test Khiops environment error"
+                if platform.system() == "Windows":
+                    temp_khiops_env_file.write("@echo off\r\n")
+                    temp_khiops_env_file.write(f"echo {test_error_message}>&2\r\n")
+                    temp_khiops_env_file.write("exit /b 1")
+                else:
+                    temp_khiops_env_file.write("#!/bin/bash\n")
+                    temp_khiops_env_file.write(f'>&2 echo "{test_error_message}"\n')
+                    temp_khiops_env_file.write("exit 1")
+
+            # Make the temporary khiops_env file executable
+            os.chmod(
+                temp_khiops_env_file_path,
+                os.stat(temp_khiops_env_file_path).st_mode | stat.S_IEXEC,
+            )
+
+            # Store initial PATH
+            original_path_env_var = os.environ["PATH"]
+
+            # Prepend path to temporary khiops_env to PATH
+            os.environ["PATH"] = (
+                temp_khiops_env_dir + os.pathsep + original_path_env_var
+            )
+
+            # Create new KhiopsLocalRunner and capture the exception
+            with self.assertRaises(KhiopsEnvironmentError) as context:
+                _ = KhiopsLocalRunner()
+
+            # Restore initial PATH
+            os.environ["PATH"] = original_path_env_var
+
+            # On Windows, restore initial KHIOPS_HOME
+            if (
+                platform.system() == "Windows"
+                and original_khiops_home_env_var is not None
+            ):
+                os.environ["KHIOPS_HOME"] = original_khiops_home_env_var
+
+        # Check that the script error message matches the expected one
+        expected_msg = (
+            "Error initializing the environment for Khiops from the "
+            f"{temp_khiops_env_file_path} script. "
+            f"Contents of stderr:\n{test_error_message}\n"
+        )
+        output_msg = str(context.exception)
+        self.assertEqual(output_msg, expected_msg)
 
     def test_runner_with_conda_based_environment(self):
         """Test that local runner works in non-Conda, Conda-based environments"""
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -231,6 +231,8 @@ jobs:
         shell: pwsh
         run: |
           $ErrorActionPreference = 'Stop'
+          Set-StrictMode -Version Latest
+          $PSNativeCommandUseErrorActionPreference = $true
 
           # Refresh environment variables by using the Chocolatey tool
           # Otherwise, the env vars set in the registry by the Khiops installer do not get updated
--- a/khiops/sklearn/estimators.py
+++ b/khiops/sklearn/estimators.py
@@ -273,8 +273,8 @@ class KhiopsEstimator(ABC, BaseEstimator):
         # pylint: disable=import-outside-toplevel
         try:
             from sklearn.utils import TransformerTags
-        except ImportError:
-            raise NotImplementedError("__sklearn_tags__ API unsupported.")
+        except ImportError as exc:
+            raise NotImplementedError("__sklearn_tags__ API unsupported.") from exc
 
         # Set the tags from _more_tags
         tags = super().__sklearn_tags__()


tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+        # Add >&2 echo "Test Khiops environment error" to the new khiops_env file
+        # Add exit 1 to the new khiops_env file
+        # Add path to the new khiops_env file (temp) dir at the beginning of
+        # os.environ["PATH"]
+        # Create KhiopsLocalRunner, which would attempt to initialize the
+        # environment from the named temp file
+        # Assert that this runner creation would raise KhiopsEnvironmentError
+        # with a message which contains the "Test Khiops environment error" string
+
+        test_error_message = "Test Khiops environment error"
```
folmos-at-orange, line 112: For better locality, this variable should be defined the closest possible to its first usage.

tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+
+        # khiops_env is always in the same directory as MODL_Coclustering
+        runner = kh.get_runner()
+        khiops_env_file_path = os.path.join(
+            os.path.dirname(runner.khiops_coclustering_path), "khiops_env"
+        )
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_khiops_env_file_path = os.path.join(temp_dir, "khiops_env")
+            if platform.system() == "Windows":
+                temp_khiops_env_file_path += ".cmd"
```
folmos-at-orange, line 125: There sohuld be a paragraph comment here: `# Create dummy khiops_env path`

tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+
+            # Copy khiops_env with execution mode bits
+            shutil.copy(khiops_env_file_path, temp_khiops_env_file_path)
```
folmos-at-orange, line 128: I'd say `# Copy the original khiops_env to set the execution mode on the dummy file`

tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+
+            # Append error code to the temp file
+            with open(temp_khiops_env_file_path, "a") as temp_khiops_env_file:
+                temp_khiops_env_file.write(f'>&2 echo "{test_error_message}"\n')
+                temp_khiops_env_file.write("exit 1")
```
folmos-at-orange, line 133: I'd say `# Replace the khiops_env a script that fails showing an error message`

tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+        # Create KhiopsLocalRunner, which would attempt to initialize the
+        # environment from the named temp file
+        # Assert that this runner creation would raise KhiopsEnvironmentError
+        # with a message which contains the "Test Khiops environment error" string
+
+        test_error_message = "Test Khiops environment error"
+
+        # Store initial PATH
+        path_env_var = os.environ["PATH"]
+
```
folmos-at-orange, line 116: - I'd do this closer to the change of `PATH`
  - For better locality
- I'd rename `original_path_env_var`
  - More precise

tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+            # Create new KhiopsLocalRunner and capture the exception
+            with self.assertRaises(KhiopsEnvironmentError) as context:
+                _ = KhiopsLocalRunner()
+
+        # Restore initial PATH
+        os.environ["PATH"] = path_env_var
+
+        # Use regex to account for potential errors in the khiops_env script
+        # proper (e.g. see issue https://github.com/KhiopsML/khiops/issues/495)
```
folmos-at-orange, line 146: I'd just say `# Check that the script error message matches the expected one`. I do not think necessary to point to an issue (should we do this for every change in Khiops core??)

khiops/core/internals/runner.py
```diff
@@ -842,7 +842,12 @@ def _initialize_khiops_environment(self):
             stderr=subprocess.PIPE,
             universal_newlines=True,
         ) as khiops_env_process:
-            stdout, _ = khiops_env_process.communicate()
+            stdout, stderr = khiops_env_process.communicate()
+            if khiops_env_process.returncode != 0:
+                raise KhiopsEnvironmentError(
+                    "Error initializing the environment for Khiops from the "
+                    f"`khiops_env` script: {stderr}"
+                )
```
folmos-at-orange, line 850: - I'd use the path of `khiops_env` instead of ``` `khiops_env` ```.
  - More useful info to diagnose problems
- At the end `... script: {stderr}` -> `... script. Contents of stderr:\n{stderr}`
  - This is to match the pattern in `_report_exit_status`.

tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+
+            # Prepend path to temporary khiops_env to PATH
+            os.environ["PATH"] = temp_dir + os.pathsep + path_env_var
+
+            # Create new KhiopsLocalRunner and capture the exception
+            with self.assertRaises(KhiopsEnvironmentError) as context:
+                _ = KhiopsLocalRunner()
+
+        # Restore initial PATH
+        os.environ["PATH"] = path_env_var
```
folmos-at-orange, line 143: Restore the path in the block where you changed it (ie. shift to the block)

folmos-at-orange (changes requested):

tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+        """Test that error is raised on reading erroneous khiops_env script"""
+        # Copy initial khiops_env to eponymous file in temporary dir
+        # Add >&2 echo "Test Khiops environment error" to the new khiops_env file
+        # Add exit 1 to the new khiops_env file
+        # Add path to the new khiops_env file (temp) dir at the beginning of
+        # os.environ["PATH"]
+        # Create KhiopsLocalRunner, which would attempt to initialize the
+        # environment from the named temp file
+        # Assert that this runner creation would raise KhiopsEnvironmentError
+        # with a message which contains the "Test Khiops environment error" string
```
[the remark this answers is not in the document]
popescu-v: Ok, to make it simple, I'll remove it.

tests/test_khiops_integrations.py
```diff
@@ -95,6 +97,61 @@ def test_runner_has_mpiexec_on_linux(self):
...
+        test_error_message = "Test Khiops environment error"
+
+        # Store initial PATH
+        path_env_var = os.environ["PATH"]
+
+        # khiops_env is always in the same directory as MODL_Coclustering
+        runner = kh.get_runner()
+        khiops_env_file_path = os.path.join(
+            os.path.dirname(runner.khiops_coclustering_path), "khiops_env"
+        )
```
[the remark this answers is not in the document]
popescu-v: Ok for the first two points.
For the last point:
- `khiops_path` holds the full path to `MODL[_openmpi|_mpich], whose `dirname` is _not_ always where `khiops_env` is (notably, on Rocky, where `MODL_openmpi` is where `mpiexec` is)
- hence, we need to construct the path to `khiops_env` by appending this file name to the directory where `MODL_Coclustering` is, because we know that, currently, `MODL_Coclustering` is precisely in the same place as `khiops_env` in all supported environments.
-

tests/test_khiops_integrations.py
popescu-v: Ok, but I'd replace "dummy" with "temporary".

tests/test_khiops_integrations.py
popescu-v: Ok, right, I could just fully replace the contents of the original script with the error code. And thus we don't need `assertRegex` anymore, as no further previous writing to `stderr` would be added.

tests/test_khiops_integrations.py
popescu-v: Ok.

tests/test_khiops_integrations.py
popescu-v: Ok.

khiops/core/internals/runner.py
popescu-v: Ok.

tests/test_khiops_integrations.py
popescu-v: Ok.

tests/test_khiops_integrations.py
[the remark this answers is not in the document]
folmos-at-orange: Ok. Then the comment should be something along the lines 
```python
# We use the Khiops CC path to find khiops_env's
# Note: The Khiops executable directory may not be the same of khiops_env
```

tests/test_khiops_integrations.py
folmos-at-orange: Ok

tests/test_khiops_integrations.py
folmos-at-orange: I like the idea of testing the output of `stderr` !

tests/test_khiops_integrations.py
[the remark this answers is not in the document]
popescu-v: Done.

tests/test_khiops_integrations.py
popescu-v: Done.

tests/test_khiops_integrations.py
popescu-v: Done.

tests/test_khiops_integrations.py
popescu-v: Done: Remove this copy altogether.

tests/test_khiops_integrations.py
popescu-v: Yes, we keep testing it, but we don't need to copy the previous script anymore. Hence, done.

tests/test_khiops_integrations.py
popescu-v: Done.

tests/test_khiops_integrations.py
popescu-v: Done.

khiops/core/internals/runner.py
popescu-v: Done.

tests/test_khiops_integrations.py
```diff
@@ -95,6 +98,86 @@ def test_runner_has_mpiexec_on_linux(self):
...
+            os.environ["PATH"] = original_path_env_var
+
+            # On Windows, restore initial KHIOPS_HOME
+            if (
+                platform.system() == "Windows"
+                and original_khiops_home_env_var is not None
+            ):
+                os.environ["KHIOPS_HOME"] = original_khiops_home_env_var
+
+        # We use the Khiops CC path to find khiops_env's
```
folmos-at-orange, line 162: - If you are using `basename`, then you can simply match `tmp_khiops_env_file_path` and eliminate all related to `khiops_env_file_path`.
- Furthermore you can match directly `tmp_khiops_env_file` without using `basename`.

folmos-at-orange (approved):

tests/test_khiops_integrations.py
popescu-v: Right, this becomes simpler, indeed. I'll add the change.

popescu-v: Done.

PR #319, opened 2024-12-19
Update default Khiops binaries version to 10.2.4

popescu-v:


folmos-at-orange (approved):
I'd rename to

`Update in CI Khiops core version to 10.2.4`

PR #344, opened 2025-02-05
Avoid pandas warning in samples when reading Accidents/Places table

### LINKED ISSUE #342: Eliminate pandas warning in samples
#### Description
Pour une démo, j'ai extrait le code d’apprentissage multi-table de la base accidents à partir de https://khiopsml.github.io/khiops-python/samples/samples_sklearn.html#samples_sklearn.khiops_classifier_multitable_snowflake vers un fichier train_sklearn.py.

En exécutant ce code, j'ai obtenu le warning suivant
```
C:\Users\Public\khiops_data\samples\Accidents\train_sklearn.py:14: DtypeWarning: Columns (2) have mixed types. Specify dtype option on import or set low_memory=False.
  places_df = pd.read_csv(os.path.join(accidents_data_dir, "Places.txt"), sep="\t")
```

Serait-il possible d'éviter ce warning intempestif?

folmos-at-orange:
Closes #342.

---


tests/test_remote_access.py
```diff
@@ -129,7 +129,8 @@ def is_in_a_conda_env():
             if not isinstance(kh.get_runner(), KhiopsLocalRunner):
                 return False
 
-            # Get path to the Khiops executable
+            # Get path to the Khiops executable (temporarily disable pylint warning)
+            # pylint: disable=protected-access
             khiops_path = kh.get_runner()._khiops_path
```
popescu-v, line 134: We need to re-enable PyLint protected access warning after this line, don't we?

folmos-at-orange: Also, this is test code, a priori `protected-access` it is overly paranoiac in this case.

popescu-v: Yes, but, IMHO, either we should systematically put these PyLint comments at the beginning of each test file where applicable, or we should put them "locally" (just before they are needed) as here, but in this case we should also cancel their effect "locally" (just before they are not needed anymore). To be as little intrusive as possible, I would go for the second choice.

folmos-at-orange: OK

popescu-v (approved):
LGTM

PR #354, opened 2025-02-07
Support float and bool targets in khiops classifiers

nairbenrekia:
_Put your message here_

---


popescu-v:
Rebase PR branch on `dev-v10`.
Change base branch of the PR to `dev-v10`.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
   - Example: 10.2.1.4 is the 5th version that supports khiops 10.2.1.
 - Internals: Changes in *Internals* sections are unlikely to be of interest for data scientists.
 
+## 10.3.0.1
```
popescu-v, line 9: Change to `# Unreleased`

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
...
+
+### Added
+
+- (`sklearn`) Supports boolean and float targets in `KhiopsClassifier`.
```
popescu-v, line 13: Change to "- (`sklearn`) Support for boolean and float targets in `KhiopsClassifier`.

tests/test_sklearn_output_types.py
```diff
@@ -71,12 +71,12 @@ def test_classifier_output_types(self):
...
         y_pred = khc.predict(X)
+        khc.fit(X_mt, y)
+        y_mt_pred = khc.predict(X_mt)
+
         y_bin = y.replace({0: 0, 1: 0, 2: 1})
         khc.fit(X, y_bin)
         y_bin_pred = khc.predict(X)
-        khc.fit(X_mt, y)
-        khc.export_report_file("report.khj")
-        y_mt_pred = khc.predict(X_mt)
```
popescu-v, line 79: Why were these changes needed, most notably, the deletion of the report file export instruction?

tests/test_sklearn_output_types.py
```diff
@@ -128,9 +142,13 @@ def test_classifier_output_types(self):
                     ),
                     "cat int": y_pred,
                     "cat string": self._replace(y_pred, {0: "se", 1: "vi", 2: "ve"}),
+                    "cat float": self._replace(y_pred, {0: 0.0, 1: 1.0, 2: 2.0}),
```
popescu-v, line 145: Simplify to something like:

```python
"cat float": self._replace(y_pred, {target: float(target) for target in (0, 1, 2)}),
```

popescu-v (changes requested):
See the comments.
Update docstring of `KhiopsClassifier.predict` with the mention of the `bool` and `float` target types: https://github.com/KhiopsML/khiops-python/pull/354/files#diff-802caefa84456ad964818a51e2c00f1decf6553480a063759a060c229fb16166R2308.

tests/test_sklearn_output_types.py
folmos-at-orange: The report export was a an unnecessary statement.

popescu-v: Ok, fair enough.

CHANGELOG.md
nairbenrekia: Ok

tests/test_sklearn_output_types.py
nairbenrekia: Moreover, I just re-ordered the other instructions.

tests/test_sklearn_output_types.py
nairbenrekia: Ok

tests/test_sklearn_output_types.py
```diff
@@ -144,6 +162,8 @@ def test_classifier_output_types(self):
                     ),
                     "cat int": y_mt_pred,
                     "cat string": self._replace(y_mt_pred, {0: "se", 1: "vi", 2: "ve"}),
+                    "cat float": self._replace(y_mt_pred, {0: 0.0, 1: 1.0, 2: 2.0}),
```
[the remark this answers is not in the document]
nairbenrekia: Ok

tests/test_sklearn_output_types.py
folmos-at-orange: > Simplify to something like:
> 
> ```python
> "cat float": self._replace(y_pred, {target: float(target) for target in (0, 1, 2)}),
> ```

Is than really simpler ?

popescu-v: At least it's more readable IMHO, as the intention is clearer IMHO.

popescu-v:
Add sklearn sample illustrating the boolean and / or floating-point targets?

tests/test_sklearn_output_types.py
folmos-at-orange: Dont agree at all but ok.

khiops/sklearn/dataset.py
```diff
@@ -738,8 +738,15 @@ def _init_target_column(self, y):
         if isinstance(y, str):
             y_checked = y
         else:
-            y_checked = column_or_1d(y, warn=True)
-
+            if hasattr(y, "dtype"):
```
folmos-at-orange, line 741: I think there is still the case of a pandas `DataFrame` with dimension 1 (not a pandas `Series`):

```python
# ...
elif hasattr(y, "dtypes"):
   y_checked = column_or_1d(y, warn=True, dtype=y.dtype.categories.dtypes[0])
```

I think these cases are missing from the tests as well, but it seems bit long for this PR. I'll make an issue.

folmos-at-orange (changes requested):

khiops/sklearn/dataset.py
nairbenrekia: True. Ok.

popescu-v (changes requested):
Update samples_sklearn.py to use boolean and / or floating-point classification targets (see relevant comment on the PR).
Check API doc building workflow for no warnings; visually inspect API docs rendering from the CI-produced artifact.

khiops/sklearn/dataset.py
```diff
@@ -738,8 +738,22 @@ def _init_target_column(self, y):
...
+                if isinstance(y.dtype, pd.CategoricalDtype):
+                    y_checked = column_or_1d(
+                        y, warn=True, dtype=y.dtype.categories.dtype
```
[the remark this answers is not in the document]
popescu-v: Yes, but we need the fallback (to calling `column_or_1d` without the `dtype` argument) in all cases where `dtype` is used (I have counted 3 of these).

On the other hand, the warning indeed only seems useful in the case that you mentioned, @folmos-at-orange. Hence, I would go for having:

- the fallback in all cases,
- the warning only if `y` contains the stringy `"True"` and `"False"`.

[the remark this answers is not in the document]
folmos-at-orange: Ok

popescu-v (changes requested):
Technically, LGTM. However, rebase of the PR branch on the latest `dev-v10` is needed before merging.

khiops/sklearn/dataset.py
popescu-v: @folmos-at-orange: Could you create the mentioned issue related to this if still applicable (especially in the light of v11), or close otherwise?

popescu-v (approved):
LGTM

khiops/samples/samples_sklearn.py
folmos-at-orange, line 1: I don't think a samples for these features are necessary. This is the kind of thing that "it should just work".

khiops/sklearn/dataset.py
```diff
@@ -430,6 +431,16 @@ def write_internal_data_table(dataframe, file_path_or_stream):
     )
 
 
+def _column_or_1d_with_dtype(y, dtype=None):
+    # 'dtype' has been introduced on `column_or_1d' since Scikit-learn 1.2;
+    if sklearn.__version__ < "1.2":
+        if pd.api.types.is_string_dtype(dtype) and y.isin(["True", "False"]).all():
+            warnings.warn("'y' contains strings of 'True' and 'False'")
```
folmos-at-orange, line 438: I'd rephrase: `y stores strings restricted to 'True'/'False' values: The predict method may return a bool vector.`

khiops/sklearn/dataset.py
folmos-at-orange: I think this is solved.

tests/test_sklearn_output_types.py
```diff
@@ -171,6 +195,7 @@ def test_classifier_output_types(self):
 
                     # Check the return type of predict
                     y_pred = khc.predict(X)
+
```
folmos-at-orange, line 198: Eliminate this blank line

folmos-at-orange, line 198: I leave this comment here because is close to the relevant code:

I think we can easily cover the case for y being a 1D DataFrame.
This can be done in another PR if necessary.
```python
        # Test for each fixture configuration
        for y_type, y in fixtures["ys"].items():
            y_type_check = fixtures["y_type_check"][y_type]
            expected_classes = fixtures["expected_classes"][y_type]
            ys = [(y, "numpy"), (pd.DataFrame({"y": y}), "pandas")]
            for dataset_type, X in fixtures["Xs"].items():
                for y_in, y_in_type in ys:
                    with self.subTest(
                        y_type=y_type,
                        dataset_type=dataset_type,
                        estimator=KhiopsClassifier.__name__,
                        y_type=y_type
                    ):
                        # Train the classifier
                        khc = KhiopsClassifier(n_trees=0)
                        khc.fit(X, y)
# ...
```

folmos-at-orange (approved):
Small things + 1 that can be done in another PR.

tests/test_sklearn_output_types.py
popescu-v: I would leave this to another PR. However, I would create an issue in this respect. @folmos-at-orange , could you initiate it?

khiops/samples/samples_sklearn.py
popescu-v: Right, but as it didn't, it's also a way of ensuring non-regression _and_ to showcase that this (user-required) feature, which was not supported, is supported from now on.

folmos-at-orange: We have the tests for this. Should we have also samples for every feature combination?

tests/test_sklearn_output_types.py
folmos-at-orange: I created #371

khiops/samples/samples_sklearn.py
popescu-v: No, but the key word in my argument is "to _show_" (i.e. [historical] users can see that this is now supported; _a contrario_, the tests are not read by the users).

tests/test_sklearn_output_types.py
popescu-v: So, we can close this conversation thread, right?

khiops/samples/samples_sklearn.py
folmos-at-orange: Again: I think this should "just work", and it is not an essential feature to showcase. Should we also add samples for 1-column dataframes as `y`, and for python value lists?. All these are accepted types in sklearn.

popescu-v: Finally, I tend to agree that dedicated samples for float and bool targets are not needed.

However, I still thing that it is important that we have an end-to-end test for this. A way of using these targets in an end-to-end test (as samples are also end-to-end tests after all) would be to add a sample which uses this feature only as part of a (realistic) problem, e.g. an Iris classification problem where the target is the answer to the question "Is Setosa?", which would involve converting all Setosa instances to `True` and all other instances to `False` in the Iris dataset. Through such a sample (which can be even just an addition to e.g. the `train_predictor_with_detect_format` sample or otherwise a variation of `deploy_model` which uses "Iris" instead of "Adult"), we would use the bool target in a sample (but as part of the sample, not as the main focus of it).

popescu-v: Finally, I tend to agree that dedicated samples for float and bool targets are not needed.

Hence, I am OK with removing these two dedicated samples; the automatic tests indeed cover this feature.

folmos-at-orange: That seems fine to me. Maybe in another PR ?

nairbenrekia: Agreed ! OK to remove them.

Issue #356, opened 2025-02-10
Create dynamically the khiops execution command from environment variables


folmos-at-orange:
#### Description
Currently, the khiops execution command line is constructed once at startup from the environment variables (usually set by `khiops-env`). Constructing it dynamically from the environment at each run would allow to change more easily the settings at each run by avoiding initializing a new runner.

popescu-v:
Depriorized to have the same status as the parent issue: it will wait for the v11 post-beta refactoring.

PR #361, opened 2025-02-19
Run initialization integration tests in conda CI

folmos-at-orange:
commit 3310037934794d8f48415ec685f6023f255f28b2 (HEAD -> 356-create-dynamically-the-khiops-execution-command-from-environment-variables, origin/356-create-dynamically-the-khiops-execution-command-from-environment-variables)
Author: Felipe Olmos <92923444+folmos-at-orange@users.noreply.github.com>
Date:   Thu Feb 20 14:20:40 2025 +0100

    Eliminate CONDA_PREFIX manipulation from runner initialization test

commit 04b202ba6157df9f9684b75692f98e220bbf8359
Author: Felipe Olmos <92923444+folmos-at-orange@users.noreply.github.com>
Date:   Wed Feb 19 15:33:28 2025 +0100

    Add runner initialization test to the conda package tests

commit c1cd407d83ecac8a77844d9606dd197e793ab162
Author: Felipe Olmos <92923444+folmos-at-orange@users.noreply.github.com>
Date:   Wed Feb 19 15:11:52 2025 +0100

    Fix minor bug in test CI

commit b8a163cd091d349fce6162813386d463689f37f3
Author: Felipe Olmos <92923444+folmos-at-orange@users.noreply.github.com>
Date:   Wed Feb 19 15:30:32 2025 +0100

    Add system-settings output dir to .gitignore


.github/workflows/conda.yml
```diff
@@ -117,7 +117,7 @@ jobs:
           PATH=$(echo $PATH | sed "s#$CONDA_PREFIX/bin:##g")
 
           # Unset *CONDA* environment variables
-          # As a corrolary, CONDA_PREFIX is unset
+          # As a corolary, CONDA_PREFIX is unset
```
popescu-v, line 120: s/corolary/corollary/

.github/workflows/conda.yml
```diff
@@ -153,18 +153,57 @@ jobs:
...
```
popescu-v, line 156: s/corolary/corollary/

.github/workflows/conda.yml
```diff
@@ -153,18 +153,57 @@ jobs:
...
           $PYTHON -m khiops.samples.samples -i deploy_coclustering -e
+          $PYTHON -m khiops.samples.samples -i train_predictor_error_handling -e
+          $PYTHON -m khiops.samples.samples_sklearn -i khiops_classifier -e
+          $PYTHON -m khiops.samples.samples_sklearn -i khiops_coclustering -e
+          $PYTHON -m khiops.samples.samples -i deploy_model -e
+          $PYTHON -m khiops.samples.samples -i deploy_coclustering -e
+      # Checkout the sources to obtain the tests
+      - name: Checkout Sources
+        uses: actions/checkout@v4
+        with:
```
popescu-v, line 176: Why not (see e.g. https://github.com/actions/checkout?tab=readme-ov-file#fetch-only-the-root-files-and-github-and-src-folder) 
```yaml
with:
  sparse-checkout: tests
  ...
```
This would allow us to get rid of the next stage, wouldn't it?

.github/workflows/conda.yml
```diff
@@ -131,17 +131,17 @@ jobs:
...
             --version ${{ inputs.khiops-samples-version || env.DEFAULT_SAMPLES_VERSION }}
-      - name: Test Conda Package on Samples
+      - name: Run Samples
         env:
           # Force > 2 CPU cores to launch mpiexec
           KHIOPS_PROC_NUMBER: 4
         run: |
-          kh-samples core -i train_predictor -e
+          kh-samples core -i deploy_model -e
+          kh-samples core -i deploy_coclustering -e
```
popescu-v, line 140: Are there possible issues that might be revealed by having a full train + deploy sample rather than a train sample only (shorter, quicker)? If not, then why update these tests?

.github/workflows/tests.yml
```diff
@@ -109,8 +109,8 @@ jobs:
             $CONDA install -y -n "$CONDA_ENV" `grep -v "^\[" khiops.egg-info/requires.txt`
             rm -rf khiops.egg-info
           done
-      - name: Prepare Tests Environment
-        if: github.ref != 'dev' && github.rev != 'dev-v10' && github.ref != 'main' && ! inputs.run-expensive-tests
+      - name: Activate Expensive Tests if Requested
```
popescu-v, line 112: IMHO the updated stage name is misleading, because what it does, is "Skip expensive tests by default, unless on the dev, dev-v10 or main Git branches". Hence, the former name was better IMHO.

tests/test_khiops_integrations.py
```diff
@@ -167,56 +167,24 @@ def test_environment_error_on_bogus_khiops_env_script(self):
...
-            # Store existing CONDA_PREFIX
-            conda_prefix = os.environ["CONDA_PREFIX"]
-
-            # Unset `CONDA_PREFIX`
-            del os.environ["CONDA_PREFIX"]
+    def test_runner_environment_initialization(self):
+        """Test that local runner initializes its environment properly"""
 
         # Create a fresh local runner
         runner = KhiopsLocalRunner()
```
popescu-v, line 174: IMHO, we don't need to create a local runner anymore. Could we just do:
```python
runner = kh.get_runner()
```
? Because we can just test that the runner that is used _by default_ sets its environment correctly, right?

.github/workflows/conda.yml
folmos-at-orange: I tried that and it didn't work as expected: it still recovered other files, so I still have to remove the files.

popescu-v (changes requested):
See the comments on the code changes.
Also:
- rename branch so that it is not linked to issue #356 anymore;
- provide a concise and to-the-point description to the PR.

.github/workflows/tests.yml
folmos-at-orange: I think the former name is not specific enough. How about "Configure Expensive Tests Setting".

tests/test_khiops_integrations.py
folmos-at-orange: That would not work, because the could be already initialized: We do not know in which context the test will be executed (with a fresh or already initialized runner). So the safest way to ensure the pre-condition is to initialize a new one.

.github/workflows/conda.yml
folmos-at-orange: I updated it because it seemed simpler to just have the deploy tasks (which include train).

.github/workflows/conda.yml
folmos-at-orange: Ok

.github/workflows/conda.yml
folmos-at-orange: Ok

.github/workflows/tests.yml
popescu-v: Ok with the new name.

tests/test_khiops_integrations.py
popescu-v: I tend to disagree: we want to test that, in any supported environment (as configured outside of the Python process), the default runner works correctly. And IMHO it should: I don't see why it wouldn't (normally a test or a test class should not leak environment changes to other tests / test classes: if this happens, then this is a bug IMHO).

.github/workflows/conda.yml
popescu-v: Ok: out of the scope of the PR, but an improvement IMHO, as it enlarges the coverage.

.github/workflows/conda.yml
folmos-at-orange: Ok

folmos-at-orange: Done

.github/workflows/conda.yml
folmos-at-orange: Done

.github/workflows/conda.yml
folmos-at-orange: Done

.github/workflows/tests.yml
folmos-at-orange: Done, and added a comment as well with your quote as well, because the logic is difficult to parse when reading.

tests/test_khiops_integrations.py
```diff
@@ -167,56 +167,24 @@ def test_environment_error_on_bogus_khiops_env_script(self):
...
 
+        .. note::
+            To test a real initialization this test should be executed alone.
+        """
         # Check that MODL* files as set in the runner exist and are executable
-        self.assertTrue(os.path.isfile(runner.khiops_path))
-        self.assertTrue(os.access(runner.khiops_path, os.X_OK))
-        self.assertTrue(os.path.isfile(runner.khiops_coclustering_path))
-        self.assertTrue(os.access(runner.khiops_coclustering_path, os.X_OK))
+        self.assertTrue(os.path.isfile(kh.get_runner().khiops_path))
```
popescu-v, line 177: I think it suffices to get the runner at the beginning of the test function and then reuse that: the default runner shouldn't change between two `assert`s, should it?

folmos-at-orange: I think we don't need to have a variable if we are accessing the default runner.

popescu-v: IMHO, the difference lies in the way the intent is marked: to me, repeated lookup emphasizes the fact that the result might change from one lookup to another, whereas a single lookup + variable assignment emphasizes the contrary, which is the case here: the runner doesn't change from one lookup to another.

folmos-at-orange: Ok

folmos-at-orange: Done.

popescu-v (approved):
LGTM.
Before merging:
- rename branch so that any relation to issue #356 is removed
- create new PR for the renamed branch
- add concise, to-the-point description to the new PR
- in the new PR's description, add link to the current (older) PR, so that we can track the discussion.

PR #381, opened 2025-04-03
380 use mpi for khiops coclustering

### LINKED ISSUE #380: Use MPI for Khiops Coclustering
#### Description
Currently, the MPI command and its arguments are concatenated only to the Khiops tool path, which ends with the `MODL` executable.

However, since `khiops-core:10.7.0-b.0` Khiops coclustering also uses MPI (currently, for model deployment, database check, and database sorting).

#### Questions/Ideas
- The goal of this issue is to concatenate the MPI command with the Khiops coclustering path as well.
  - That is, the MPI command (with arguments) should be prepended to both Khiops and Khiops coclustering. Hence, the test on the Khiops tool name is no longer necessary when prepending the MPI command (with arguments) to the Khiops tool command.

popescu-v:
closes #380 

---

### DIFF
--- a/tests/test_khiops_integrations.py
+++ b/tests/test_khiops_integrations.py
@@ -4,7 +4,7 @@
 # which is available at https://spdx.org/licenses/BSD-3-Clause-Clear.html or         #
 # see the "LICENSE.md" file for more details.                                        #
 ######################################################################################
-"""Tests for executing fit multiple times on multi-table data"""
+"""Various integration tests"""
 
 import os
 import platform
@@ -13,8 +13,10 @@ import stat
 import subprocess
 import tempfile
 import unittest
+from unittest.mock import MagicMock, patch
 
 import khiops.core as kh
+import khiops.core.internals.filesystems as fs
 from khiops.core.exceptions import KhiopsEnvironmentError
 from khiops.core.internals.runner import KhiopsLocalRunner
 from khiops.extras.docker import KhiopsDockerRunner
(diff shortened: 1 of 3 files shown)


tests/test_khiops_integrations.py
```diff
@@ -208,6 +211,52 @@ def test_runner_environment_initialization(self):
...
+        # Get current runner
+        runner = kh.get_runner()
+
+        # Get path to the Iris dataset
+        iris_data_dir = fs.get_child_path(runner.samples_dir, "Iris")
+
+        quote = '"' if platform.system() == "Windows" else "'"
+
+        # Run Khiops through an API function and test its stdout
+        with contextlib.redirect_stdout(io.StringIO()) as khiops_stdout:
```
folmos-at-orange, line 226: This needs a better explanation for a future reader
- Run an API function _with trace enabled_, this prints the execution command to the stdout
- You capture the stdout and check that the execution line matches the mpi command args in the runner.

Having said that, I find something off about this test, because we print the trace just after setting `mpi_command_args`. So we are basically checking that we print the trace well, not that we execute the process with the correct parameters.

I think it'd be if the test does the following
- Mock `subprocess.Popen`
- Capture the arguments, the first part should match `mpi_command_args`

popescu-v: Right, but enabling the trace entails printing `khiops_call`, which is basically the same as `khiops_process_args` which is passed to `Popen`, modulo some quoting. `khiops_call` preserves the relevant piece of information that we are after: that the MPI command is prepended to the Khiops command. Hence, I'm not sure we need to mock `Popen`, as long as, for the purpose of the current test, the trace should suffice.

tests/test_khiops_integrations.py
```diff
@@ -208,6 +211,52 @@ def test_runner_environment_initialization(self):
...
+                "Iris",
+                fs.get_child_path(iris_data_dir, "Iris.txt"),
+                trace=True,
+            )
+        expected_khiops_command = (
+            f"{' '.join(runner.mpi_command_args)} {runner.khiops_path}"
+        )
+        self.assertTrue(
+            expected_khiops_command in khiops_stdout.getvalue().replace(quote, "")
+        )
```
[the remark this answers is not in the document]
popescu-v: Done.

tests/test_khiops_integrations.py
```diff
@@ -208,6 +211,52 @@ def test_runner_environment_initialization(self):
...
+    def test_khiops_and_khiops_coclustering_are_run_with_mpi(self):
+        """Test that MODL and MODL_Coclustering are run with MPI"""
+
+        # Get current runner
+        runner = kh.get_runner()
+
+        # Get path to the Iris dataset
+        iris_data_dir = fs.get_child_path(runner.samples_dir, "Iris")
+
+        quote = '"' if platform.system() == "Windows" else "'"
```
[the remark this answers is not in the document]
popescu-v: Done.

tests/test_khiops_integrations.py
popescu-v: Enhanced the comments though.

folmos-at-orange: Besides testing the effective call arguments (those received by Popen) are the expected ones, it'd be cleaner with a mock as you don't need to resort to `trace` or string matching.

```python
import unittest
from unittest.mock import MagicMock, patch
from khiops import core as kh

class TestSuite(unittest.TestCase):
    def test(self):
        # Get the runner
        runner = kh.get_runner()
        
        # Obtain the version just to set up the runner
        _ = kh.get_khiops_version()

        # Create the mock
        mock_popen = MagicMock()
        mock_popen.return_value.__enter__.return_value.communicate.return_value = (b'', b'')
        mock_popen.return_value.__enter__.return_value.returncode = 0

        # Call `check_database` with the mock
        with patch('subprocess.Popen', mock_popen):
            kh.check_database(
                f"{kh.get_samples_dir()}/Iris.kdic",
                "Iris",
                f"{kh.get_samples_dir()}/Iris.txt",
            )

        print(mock_popen.call_args) 
        # Check that runner._mpi_command_args match the first components of `call_args`
        #...
```

popescu-v: Ok, seems clearer, more focused and less brittle, indeed.

popescu-v: Done.

folmos-at-orange (approved):

PR #402, opened 2025-04-25
Detect "unhappy" installation states

tramora:
- Warn when the version tuple (major, minor, patch) of Khiops does not match the Khiops Python library one
- Fix the types of the returned objects in `_build_status_message`
- Detect when the library is installed by something else than conda in a conda environment
 - Detect when the conda execution environment does not match the installation one

---

- [ X] Rebase your branch to the latest version of `dev` (or `main` for release PRs)


khiops/core/internals/runner.py
```diff
@@ -397,46 +432,86 @@ def _build_status_message(self):
...
+                error = (
+                    f"Khiops Python library installation path '{package_dir}' "
+                    f"does not match the current Conda environment "
+                    f"'{os.environ['CONDA_PREFIX']}'. "
+                    f"Please install the Khiops Python library "
+                    f"in the current Conda environment.\n"
+                )
+                errors_list.append(error)
+            # Ensure no mix between conda and pip exists within a conda environment
+            current_module_installer = _get_current_module_installer()
```
popescu-v, line 472: How about running a genuine binary+Pip Khiops installation from within an (unrelated) Conda environment? This should not be "unhappy" IMHO.

popescu-v:
Perhaps perform a 3-times check of "happy path":
1. Check that Python executable path matches khiops `__file__`;
2. If so, then check that khiops `__file__` also matches (relevant part of) `CONDA_PREFIX`;
3. If so, then check that Khiops executable path matches (relevant part of) `CONDA_PREFIX` ?

Any `else` is an unhappy path.

khiops/core/internals/runner.py
```diff
@@ -179,7 +180,18 @@ def _check_conda_env_bin_dir(conda_env_bin_dir):
...
 
 def _infer_khiops_installation_method(trace=False):
-    """Return the Khiops installation method"""
+    """Return the Khiops installation method
+
+    Definitions :
+    - 'conda' environment will contain binaries, shared libraries and python modules
+    - 'conda-based' environment is quite similar to 'conda' except that
+       it will not be activated previously nor during the execution
+       and thus the CONDA_PREFIX environment variable will remain undefined
```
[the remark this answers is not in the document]
tramora: The comment was amended

khiops/core/internals/runner.py
tramora: I took as granted the requirement of not mixing the installation modes (installing via pip for example in a conda env) but it seems that this is not the real issue because this case works. I removed this false assumption

khiops/core/internals/runner.py
```diff
@@ -397,46 +432,86 @@ def _build_status_message(self):
...
+                    f"Khiops Python library installation path '{package_dir}' "
+                    f"does not match the current Conda environment "
+                    f"'{os.environ['CONDA_PREFIX']}'. "
+                    f"Please install the Khiops Python library "
+                    f"in the current Conda environment.\n"
+                )
+                errors_list.append(error)
+            # Ensure no mix between conda and pip exists within a conda environment
+            current_module_installer = _get_current_module_installer()
+            if current_module_installer != "conda":
```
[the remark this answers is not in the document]
tramora: The new version unsures now that the khiops binary is in the same (activated) conda environment as the khiops-python library.

tramora:
> Perhaps perform a 3-times check of "happy path":
> 
>     1. Check that Python executable path matches khiops `__file__`;

After a few attempts it seems that this check is pointless and a big hassle. Pointless because if a wrong python interpreter is used it wouldn't find the correct library (and digging programmatically from the python library for the python executable will lead to the sys.executable). A big hassle because of the OS specificities ('Scripts' folder for Windows instead of 'bin' for example)

> 
>     2. If so, then check that khiops `__file__` also matches (relevant part of) `CONDA_PREFIX` if defined;
> 

Indeed. This check is performed.

>     3. If so, then check that Khiops executable path matches (relevant part of) `CONDA_PREFIX` ?
> 
Indeed. This check is performed.
> 
> Any `else` is an unhappy path, if `CONDA_PREFIX` is defined.

popescu-v:
> > Perhaps perform a 3-times check of "happy path":
> > ```
> > 1. Check that Python executable path matches khiops `__file__`;
> > ```
> 
> After a few attempts it seems that this check is pointless and a big hassle. Pointless because if a wrong python interpreter is used it wouldn't find the correct library (and digging programmatically from the python library for the python executable will lead to the sys.executable). A big hassle because of the OS specificities ('Scripts' folder for Windows instead of 'bin' for example)

Wouldn't this accommodate both Windows and Linux: `khiops.__file__.startswith(str(pathlib.Path(sys.executable).parents[1]))`?
Because otherwise, how do we make sure, _if not in a Conda env_, that we don't launch a Virtualenv-based Python which imports a system-wide installed Khiops?

> 
> > ```
> > 2. If so, then check that khiops `__file__` also matches (relevant part of) `CONDA_PREFIX` if defined;
> > ```
> 
> Indeed. This check is performed.
> 
> > ```
> > 3. If so, then check that Khiops executable path matches (relevant part of) `CONDA_PREFIX` ?
> > ```
> 
> Indeed. This check is performed.
> 
> > Any `else` is an unhappy path, if `CONDA_PREFIX` is defined.

khiops/core/internals/runner.py
```diff
@@ -218,6 +254,116 @@ def _check_executable(bin_path):
...
+                f"but with '{current_library_installer}' installer. "
+                "This will probably lead to unexpected errors. "
+                "Go to https://khiops.org for instructions to re-install it.\n",
+                "",  # category - useless in our case
+                "",  # filename - useless in our case
+                0,  # lineno - useless in our case
+            )
+            warnings_list.append(warning)
+
+        # let's consider only the 'binary+pip' and 'conda-based' installations here
```
popescu-v, line 355: Add check that, when not in an activated Conda environment, `khiops.__file__.startswith(str(pathlib.Path(sys.executable).parents[1]))` is `True`? Perhaps to modulate the `.parents` index according to the OS type?

.github/workflows/api-docs.yml
```diff
@@ -53,7 +53,11 @@ jobs:
       - name: Add pip scripts directory to path
         run: echo PATH="$PATH:/github/home/.local/bin" >> "$GITHUB_ENV"
       - name: Install doc build requirements
+        shell: bash
         run: |
+          python -m venv .venvdocs
+          source .venvdocs/bin/activate
```
popescu-v, line 59: I would use the POSIX-standard `.` instead of the Bash-specific `source`.

khiops/core/internals/runner.py
```diff
@@ -129,26 +131,71 @@ def _khiops_env_file_exists(env_dir):
...
+
+
+def _infer_base_dir_for_conda_based_or_pip_installations():
+    """Infer reference directory (base directory)
+     for Khiops installations
+
+    This function can detect
+    - 'conda' and 'conda-based' installations
+    - system-wide pure python installation (in a dist-packages folder)
+    - pure python virtual environment (in a site-packages folder)
```
popescu-v, line 160: Add trailing "installation".

khiops/core/internals/runner.py
```diff
@@ -129,26 +131,71 @@ def _khiops_env_file_exists(env_dir):
...
+
+    :returns base_dir : Path | str,
+    base_dir is an empty string if a borderline installation is detected
+    """
     assert os.path.basename(Path(__file__).parents[2]) == "khiops", (
         f"The {os.path.basename(__file__)} file has been moved. "
         "Please fix the `Path.parents` in this method "
-        "so it finds the conda environment directory of this module"
+        "so it finds the conda environment directory of this library"
```
popescu-v, line 168: I would keep "module", because:
- this is an `assert` message, whence addressed to the developers of the Khiops Python library;
- strictly speaking `__file` refers to the current _module_, as opposed to the library, which refers to the package itself.

khiops/core/internals/runner.py
```diff
@@ -129,26 +131,71 @@ def _khiops_env_file_exists(env_dir):
...
+    # or {virtual env root}\Lib\site-packages\khiops\core\internals\runner.py
     if platform.system() == "Windows":
-        conda_env_dir = current_file_path.parents[5]
-    # Linux/macOS:
-    # Match $CONDA_PREFIX/[Ll]ib/python3.X/site-packages/khiops/core/internals/runner.py
+        # safeguard to prevent an IndexError on borderline installations
+        if len(current_file_path.parents) < 6:
+            base_dir = ""
+        else:
+            base_dir = current_file_path.parents[5]
```
popescu-v, line 184: I would keep `base_dir`'s type consistent and stringify it here.

khiops/core/internals/runner.py
```diff
@@ -129,26 +131,71 @@ def _khiops_env_file_exists(env_dir):
...
+    # or {virtual env root}/
+    #       [Ll]ib/python3.X/site-packages/khiops/core/internals/runner.py
     else:
-        conda_env_dir = current_file_path.parents[6]
-    env_bin_dir = os.path.join(str(conda_env_dir), "bin")
+        # safeguard to prevent an IndexError on borderline installations
+        if len(current_file_path.parents) < 7:
+            base_dir = ""
+        else:
+            base_dir = current_file_path.parents[6]
```
popescu-v, line 196: I would keep base_dir's type consistent and stringify it here.

khiops/core/internals/runner.py
```diff
@@ -218,6 +278,153 @@ def _check_executable(bin_path):
...
+    try:
+        # Each time a python library is installed a 'dist-info' folder is created
+        # Normalized files can be found in this folder
+        installer_files = [path for path in files("khiops") if path.name == "INSTALLER"]
+        if len(installer_files) > 0:
+            return installer_files[0].read_text().strip()
+    except PackageNotFoundError:
+        # The python library is not installed via standard tools like conda, pip...
+        pass
+    return "unknown"
```
popescu-v, line 300: Why not put `return "unknown"` inside the `except PackageNotFoundError` block?

khiops/core/internals/runner.py
```diff
@@ -218,6 +278,153 @@ def _check_executable(bin_path):
...
+            # must match the library installation
+            if not (
+                Path(sys.executable).parents[1] == base_dir
+                or (
+                    # Under Windows, python is not in a bin/ folder
+                    # for conda-based installations
+                    platform.system() == "Windows"
+                    and Path(sys.executable).parents[0] == base_dir
+                )
+            ):
```
popescu-v, line 417: I would simplify this snippet, by relying on the `and` operator having a higher precendence than the `or` operator, thus:
```python
if (platform.system() == "Windows" and Path(sys.executable).parents[0] != base_dir
    or Path(sys.executable).parents[1] != base_dir
):
    error = ...
```

khiops/core/internals/runner.py
```diff
@@ -955,21 +1186,28 @@ def _initialize_khiops_version(self):
 
         self._khiops_version = KhiopsVersion(khiops_version_str)
 
-        # Warn if the khiops version is too far from the Khiops Python library version
+        # Warn if the khiops version does not match the Khiops Python library version
+        # Currently the check is very strict
+        # (major.minor.patch must be the same), it could be relaxed later
```
popescu-v, line 1191: I would drop the `, it could be relaxed later` part, as I'm not aware of any specification along these lines. For the time being, all we know is that the check is on the `major.minor.patch`.

khiops/core/internals/runner.py
```diff
@@ -955,21 +1186,28 @@ def _initialize_khiops_version(self):
...
         compatible_khiops_version = khiops.get_compatible_khiops_version()
-        if self._khiops_version.major > compatible_khiops_version.major:
+        if (
+            (self._khiops_version.major != compatible_khiops_version.major)
+            or (self._khiops_version.minor != compatible_khiops_version.minor)
```
popescu-v, line 1195: I would precede this with a comment like: `# KhiopsVersion implements an equality operator, which however also takes pre-release tags into account. The restriction here does not apply to pre-release tags.`

popescu-v (changes requested):
See pending comments.

Also, I would put the `api-docs.yml` changes in a separate commit.

tramora:
> See pending comments.
> 
> Also, I would put the `api-docs.yml` changes in a separate commit.

These changes were put in a different commit

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -10,7 +10,7 @@ ARG KHIOPS_REVISION
 RUN true \
   # Install git (for khiops-python version calculation) and pip \
   && apt-get -y update \
-  && apt-get -y --no-install-recommends install git python3-pip zip pandoc wget ruby-dev \
+  && apt-get -y --no-install-recommends install git python3-pip python3.10-venv zip pandoc wget ruby-dev \
```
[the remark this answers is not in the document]
tramora: A long comment explains the motivation of this installation

khiops/core/internals/runner.py
tramora: I have no added yet the comment because I'm not sure to fully understand it

khiops/core/internals/runner.py
tramora: I've simplified the expression

khiops/core/internals/runner.py
tramora: It is important to keep this statement outside the except block because it is possible that a dist-info folder exists but no "INSTALLER" file cannot be found and thus no installation type cannot be returned

khiops/core/internals/runner.py
tramora: Modified

khiops/core/internals/runner.py
```diff
@@ -129,26 +131,71 @@ def _khiops_env_file_exists(env_dir):
...
+def _infer_base_dir_for_conda_based_or_pip_installations():
+    """Infer reference directory (base directory)
+     for Khiops installations
+
+    This function can detect
+    - 'conda' and 'conda-based' installations
+    - system-wide pure python installation (in a dist-packages folder)
+    - pure python virtual environment (in a site-packages folder)
+
+    :returns base_dir : Path | str,
```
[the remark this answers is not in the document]
tramora: Modified

khiops/core/internals/runner.py

khiops/core/internals/runner.py
popescu-v: Right, you would have to duplicate the `return "unknown"` inside the `try` block, after the `if` block; but we could drop the `pass`:
```python
try:
    # Each time a python library is installed a 'dist-info' folder is created
    # Normalized files can be found in this folder
    installer_files = [path for path in files("khiops") if path.name == "INSTALLER"]
    if len(installer_files) > 0:
        return installer_files[0].read_text().strip()

    # >No "INSTALLER" file is found inside the package metadata
    return "unknown"
except PackageNotFoundError:
    # The python library is not installed via standard tools like conda, pip...
    return "unknown"
```
This seems more readable to me, especially with a comment added (see above the comment which starts with ">").

packaging/docker/khiopspydev/Dockerfile.ubuntu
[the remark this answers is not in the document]
popescu-v: Ok, got it, thanks.

khiops/core/internals/runner.py
tramora: The code block was modified

khiops/core/internals/runner.py
tramora: I added the comment

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,153 @@ def _check_executable(bin_path):
...
+                "",  # category - useless in our case
+                "",  # filename - useless in our case
+                0,  # lineno - useless in our case
+            )
+            warnings_list.append(warning)
+
+        # the conda environment must match the library installation
+        # this check may be superfluous because a mismatch is highly improbable
+        # unless the environment variable was modified in purpose
+        if not library_root_dir.as_posix().startswith(os.environ["CONDA_PREFIX"]):
```
popescu-v, line 365: Replace `library_root_dir.as_posix()` with `str(library_root_dir)`. `as_posix` risks not doing exactly the same thing as `str`, especially on Windows: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,153 @@ def _check_executable(bin_path):
...
+        # - system-wide 'binary+pip' installation (without virtual env)...
+        # - 'binary+pip' installation under a virtual env
+        # (an empty string means a borderline installation was found,
+        # no further check cannot be performed)
+        base_dir = _infer_base_dir_for_conda_based_or_pip_installations()
+        if len(base_dir) > 0:
+            # the python executable location
+            # (within the virtual env or the conda-based env)
+            # must match the library installation
+            if (
```
popescu-v, line 420: This works for a virtualenv, but not for system-wide installations. Hence:
- issue this error if this supplemental condition is met `sys.base_prefix != sys.prefix` (we are in a virtualenv)
- else, issue this error if `not library_root_dir.startswith(sys.base_prefix)`

tramora: Indeed the use sys.base_prefix and sys.prefix is the simplest way to detect if a virtualenv is activated.
I added an error case worth considering in the system-wide case : when the python executable is not really the one installed system-wide (easily simulated by `ln -s /usr/bin/python /tmp/python` for example)

khiops/core/internals/runner.py

.github/workflows/api-docs.yml
```diff
@@ -65,6 +69,7 @@ jobs:
       # Clone the Khiops Python tutorial repository while building the documentation
       - name: Build Sphinx Documentation
         run: |
+          . .venvdocs/bin/activate
```
popescu-v, line 72: I'm not sure we need to use a virtual environment here (see comment above).

khiops/core/internals/runner.py
```diff
@@ -158,6 +153,54 @@ def _infer_env_bin_dir_for_conda_based_installations():
     return env_bin_dir
 
 
+def _infer_base_dir_for_conda_based_or_pip_installations():
+    """Infer reference directory (base directory)
```
popescu-v, line 157: Use standard docstring "template" (see Core API functions for example).

khiops/core/internals/runner.py
```diff
@@ -158,6 +153,54 @@ def _infer_env_bin_dir_for_conda_based_installations():
...
+    This function can detect
+    - 'conda' and 'conda-based' installations
+    - system-wide pure python installation (in a dist-packages folder)
+    - pure python virtual environment installation (in a site-packages folder)
+
+    :returns base_dir : str,
+    base_dir is an empty string if a borderline installation is detected
+    """
+    assert os.path.basename(Path(__file__).parents[2]) == "khiops", (
+        f"The {os.path.basename(__file__)} file has been moved. "
```
popescu-v, line 169: I'm not sure this first line is still necessary. I would keep the others though.

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+    """
+
+    try:
+        # Each time a python library is installed a 'dist-info' folder is created
+        # Normalized files can be found in this folder
+        installer_files = [path for path in files("khiops") if path.name == "INSTALLER"]
+        if len(installer_files) > 0:
+            return installer_files[0].read_text().strip()
+
+        # No "INSTALLER" file is found inside the package metadata
```
popescu-v, line 309: How can this happen? Non-standard (obsolete?) packaging?

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+        return "unknown"
+    except PackageNotFoundError:
+        # The python library is not installed via standard tools like conda, pip...
+        return "unknown"
+
+
+def _detect_library_installation_incompatibilities(
```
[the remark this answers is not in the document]
popescu-v: I would make this a private method of the `KhiopsLocalRunner` (I think; or upper in the class hierarchy if needed), rather than passing the runner as the first argument.

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+    library_root_dir : PosixPath
+        path to this current library
+
+    errors_list : list
+        Collection of errors messages (str),
+        optionally modified during the call.
+
+    warnings_list : list
+        Collection of warning messages (WarningMessage),
+        optionally modified during the call.
```
popescu-v, line 340: I'd remove these from the parameters list, and build them inside the method and return them.

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+    """Detect known incompatible installations of this python library
+    in the 3 installation modes see `_infer_khiops_installation_method`
+    (binary+pip, conda, conda-based)
+
+    The errors_list or warning_list collections
+    are populated if an issue is detected
+
+    Parameters
+    ----------
+    runner : KhiopsRunner
```
popescu-v, line 328: I'd make this a private method of the `KhiopsLocalRunner` (or upper in the class hierarchy if needed).

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+
+    errors_list : list
+        Collection of errors messages (str),
+        optionally modified during the call.
+
+    warnings_list : list
+        Collection of warning messages (WarningMessage),
+        optionally modified during the call.
+    """
+    if "CONDA_PREFIX" in os.environ:
```
popescu-v, line 342: I'd add `and os.path.join(os.environ["CONDA_PREFIX"], "bin")) in os.environ["PATH"]` to the `if`, because these 2 are the invariants of a "proper" Conda environment AFAIK.

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+        # activated 'conda' installation
+
+        # unfortunately, under windows the system-wide msmpi
+        # will take precedence over the one installed
+        # in the current 'conda' environment
```
popescu-v, line 347: I would remove or change this comment to reflect the fact that this is just a warning that can be ignored.

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+        windows_msmpi_path = os.path.join("c:", "Windows", "System32", "msmpi.dll")
+        if platform.system() == "Windows" and os.path.exists(windows_msmpi_path):
+            warning = warnings.WarningMessage(
```
popescu-v, line 350: Why do we need to use `WarningMessage` objects here, as only the string (obtained through the `.message` attribute) is used anyway? I would keep the previous approach of using strings for warning as well.

khiops/core/internals/runner.py
```diff
@@ -412,46 +644,70 @@ def _build_status_message(self):
         Returns
         -------
         tuple
-            A 2-tuple containing:
+            A 3-tuple containing in this order :
             - The status message
-            - A list of warning messages
+            - A list of error messages (str)
+            - A list of warning messages (WarningMessage)
```
popescu-v, line 650: I'd keep using `str` for the warnings, as before (see comment above in this respect).

khiops/core/internals/runner.py
```diff
@@ -412,46 +644,70 @@ def _build_status_message(self):
...
+            print("Warnings:")
+            for warning in warnings_list:
+                print(f"\tWarning: {warning.message}\n")
+
+        if len(errors_list) == 0:
+            return 0
         else:
-            print("")
-        return 0
+            return 1
```
popescu-v, line 710: I'd remove the `else` and unindent the `return 1`

khiops/core/internals/runner.py
```diff
@@ -970,21 +1226,31 @@ def _initialize_khiops_version(self):
...
         compatible_khiops_version = khiops.get_compatible_khiops_version()
-        if self._khiops_version.major > compatible_khiops_version.major:
+        # KhiopsVersion implements the equality operator, which however also
+        # takes pre-release tags into account.
+        # The restriction here does not apply to pre-release tags
+        if (
+            (self._khiops_version.major != compatible_khiops_version.major)
+            or (self._khiops_version.minor != compatible_khiops_version.minor)
+            or (self._khiops_version.patch != compatible_khiops_version.patch)
+        ):
```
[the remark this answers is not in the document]
tramora: test changed

khiops/core/internals/runner.py
tramora: fixed

khiops/core/internals/runner.py
tramora: The comment was updated

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+                "does not match the current Conda environment "
+                f"'{os.environ['CONDA_PREFIX']}'. "
+                "Please install the Khiops Python library "
+                "in the current Conda environment. "
+                "Go to https://khiops.org for instructions.\n"
+            )
+            errors_list.append(error)
+        # the khiops executable path must also match the conda environment one
+        # meaning khiops core was installed using conda
+        if not os.path.dirname(runner.khiops_path).startswith(
```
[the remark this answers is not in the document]
tramora: simplified

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+                f"You can ignore this warning.\n",
+                "",  # category - useless in our case
+                "",  # filename - useless in our case
+                0,  # lineno - useless in our case
+            )
+            warnings_list.append(warning)
+
+        # the conda environment must match the library installation
+        # this check may be superfluous because a mismatch is highly improbable
+        # unless the environment variable was modified in purpose
```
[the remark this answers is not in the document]
tramora: comment removed

khiops/core/internals/runner.py
tramora: Message corrected

khiops/core/internals/runner.py
```diff
@@ -129,23 +131,16 @@ def _khiops_env_file_exists(env_dir):
...
+     this function will return 'bin'
 
-    # Windows: Match %CONDA_PREFIX%\Lib\site-packages\khiops\core\internals\runner.py
-    if platform.system() == "Windows":
-        conda_env_dir = current_file_path.parents[5]
-    # Linux/macOS:
-    # Match $CONDA_PREFIX/[Ll]ib/python3.X/site-packages/khiops/core/internals/runner.py
-    else:
-        conda_env_dir = current_file_path.parents[6]
+    :return: absolute path (str) of the 'bin' dir
```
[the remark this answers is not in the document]
tramora: corrected

khiops/core/internals/runner.py
tramora: corrected

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
         )
 
 
+def _get_current_library_installer():
+    """Tells how the python library was installed
```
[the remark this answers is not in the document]
tramora: Fixed

khiops/core/internals/runner.py
```diff
@@ -412,46 +644,70 @@ def _build_status_message(self):
...
         status_msg += f"runner class        : {self.__class__.__name__}\n"
         status_msg += f"root temp dir       : {self.root_temp_dir}\n"
         status_msg += f"sample datasets dir : {samples_dir_path}\n"
-        status_msg += f"package dir         : {Path(__file__).parents[2]}\n"
-        return status_msg, warning_list
+        status_msg += f"library root dir         : {library_root_dir}\n"
+
+        errors_list = []
+
+        _detect_library_installation_incompatibilities(
```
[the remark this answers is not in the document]
tramora: Modification done

khiops/core/internals/runner.py
tramora: Modification done

khiops/core/internals/runner.py
tramora: Modified

.github/workflows/api-docs.yml
```diff
@@ -54,6 +54,10 @@ jobs:
         run: echo PATH="$PATH:/github/home/.local/bin" >> "$GITHUB_ENV"
       - name: Install doc build requirements
         run: |
+          # Build the doc inside a virtual environment to make consistent use of dependencies
+          python -m venv .venvdocs
```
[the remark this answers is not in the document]
popescu-v: One can also use `--user`: `python -m pip install --user ...`

khiops/core/internals/runner.py
```diff
@@ -129,23 +131,19 @@ def _khiops_env_file_exists(env_dir):
...
+        where the khiops binaries are installed
 
-    # Windows: Match %CONDA_PREFIX%\Lib\site-packages\khiops\core\internals\runner.py
-    if platform.system() == "Windows":
-        conda_env_dir = current_file_path.parents[5]
-    # Linux/macOS:
-    # Match $CONDA_PREFIX/[Ll]ib/python3.X/site-packages/khiops/core/internals/runner.py
-    else:
-        conda_env_dir = current_file_path.parents[6]
+        Borderline case : if no Conda-based Khiops installation is found
```
popescu-v, line 143: Precede this with `.. note:: ` (see e.g. https://github.com/KhiopsML/khiops-python/blob/b91e295297034626d34935f9ba89a1659afd4ec5/khiops/core/api.py#L156).

khiops/core/internals/runner.py
```diff
@@ -412,46 +496,67 @@ def _build_status_message(self):
...
             samples_dir_path = self.samples_dir
         if caught_warnings is not None:
-            warning_list += caught_warnings
+            # caught_warnings contains a list of WarningMessage
+            warnings_list.extend([w.message for w in caught_warnings])
+
+        # the following path is accurate only if the current file
+        # is still in the 'khiops.core.internals' package
+        assert os.path.basename(Path(__file__).parents[2]) == "khiops", (
+            f"The {os.path.basename(__file__)} file has been moved. "
```
popescu-v, line 515: Update here as well, by removing the "has been moved part".

khiops/core/internals/runner.py
```diff
@@ -970,21 +1075,196 @@ def _initialize_khiops_version(self):
...
+        Returns
+        -------
+        errors_list : list
+            Collection of the new errors messages (str)
+
+        warnings_list : list
+            Collection of the new warning messages (str)
+        """
+
+        errors_list = warnings_list = []
```
popescu-v, line 1128: No, because both lists get updated as soon as we append to one of them. They must be initialized separately.

khiops/core/internals/runner.py
```diff
@@ -233,6 +289,182 @@ def _check_executable(bin_path):
...
+            errors_list.append(error)
+    else:
+        # 'binary+pip', 'conda-based' or borderline installations
+
+        # ensure a known installer was used otherwise unexpected issues can occur
+        # if the installer is unknown we face a borderline installation
+        # (for example a run inside a cloned git repo)
+        current_library_installer = _get_current_library_installer()
+        if current_library_installer not in ("conda", "pip"):
+            warning = warnings.WarningMessage(
```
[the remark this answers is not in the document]
popescu-v: Ok.

khiops/core/internals/runner.py
```diff
@@ -412,46 +496,67 @@ def _build_status_message(self):
...
+
+        # Print the warnings (if any)
+        if warnings_list:
+            print("Warnings:")
+            for warning in warnings_list:
+                print(f"\tWarning: {warning}\n")
+
+        if len(errors_list) == 0:
+            return 0
+
```
popescu-v, line 558: Remove this line.

khiops/core/internals/runner.py
```diff
@@ -412,46 +496,67 @@ def _build_status_message(self):
...
+            print("Errors to be fixed:")
+            for error in errors_list:
+                print(f"\tError: {error}\n")
+
+        # Print the warnings (if any)
+        if warnings_list:
+            print("Warnings:")
+            for warning in warnings_list:
+                print(f"\tWarning: {warning}\n")
+
```
popescu-v, line 555: Either remove this empty line or (better IMHO) make it a paragraph by adding a comment.

popescu-v (changes requested):
There are still a few pending changes (see comments), among which a needed fix.

khiops/core/internals/runner.py
tramora: I added a comment

khiops/core/internals/runner.py
tramora: Modified

khiops/core/internals/runner.py
```diff
@@ -970,21 +1074,197 @@ def _initialize_khiops_version(self):
...
+        ----------
+        library_root_dir : PosixPath
+            path to this current library
+
+
+        Returns
+        -------
+        list
+            Collection of the new errors messages (str)
+
```
popescu-v, line 1122: Should be (cf. e.g. https://github.com/KhiopsML/khiops-python/blob/e4ebcdc0e4ed9b38b61ba3e23f323a73466dde3a/khiops/core/api.py#L798):
```python
Returns
--------
tuple
    A 2-tuple containing:
        - a list of error messages
        - a list of warning messages
```

popescu-v:
Also, add support for `pip install --user` (intermediate between global, system-wide install, and in-virtualenv install).

khiops/core/internals/runner.py
tramora: docstring fixed

folmos-at-orange (changes requested):

khiops/core/internals/runner.py
```diff
@@ -129,23 +132,19 @@ def _khiops_env_file_exists(env_dir):
 
 
 def _infer_env_bin_dir_for_conda_based_installations():
-    """Infer reference directory for Conda-based Khiops installations"""
-    assert os.path.basename(Path(__file__).parents[2]) == "khiops", (
-        f"The {os.path.basename(__file__)} file has been moved. "
-        "Please fix the `Path.parents` in this method "
-        "so it finds the conda environment directory of this module"
-    )
+    """Infer the bin directory of
```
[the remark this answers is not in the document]
tramora: Fixed

khiops/core/internals/runner.py
```diff
@@ -129,23 +132,19 @@ def _khiops_env_file_exists(env_dir):
...
+    *supposed* Conda-based Khiops installations
 
-    # Obtain the full path of the current file
-    current_file_path = Path(__file__).resolve()
+    Returns
+    -------
+    str
+        absolute path of the 'bin' dir
+        where the khiops binaries are installed
```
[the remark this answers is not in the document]
tramora: Fixed

khiops/core/internals/runner.py
```diff
@@ -158,6 +157,56 @@ def _infer_env_bin_dir_for_conda_based_installations():
     return env_bin_dir
 
 
+def _infer_base_dir_for_conda_based_or_pip_installations():
+    """Infer reference directory (base directory)
```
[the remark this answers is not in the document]
tramora: Fixed

khiops/core/internals/runner.py
```diff
@@ -158,6 +157,56 @@ def _infer_env_bin_dir_for_conda_based_installations():
...
+     for Khiops installations
+
+    This function can detect
+    - 'conda' and 'conda-based' installations
+    - system-wide pure python installation (in a dist-packages folder)
+    - pure python virtual environment installation (in a site-packages folder)
+
```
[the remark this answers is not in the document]
tramora: The line was added

khiops/core/internals/runner.py
```diff
@@ -158,6 +157,56 @@ def _infer_env_bin_dir_for_conda_based_installations():
...
+    Returns
+    -------
+    str
+        base_dir
```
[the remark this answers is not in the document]
tramora: Replaced by what is returned and the borderline cas was prefixed with `.. note:`

khiops/core/internals/runner.py
```diff
@@ -158,6 +157,56 @@ def _infer_env_bin_dir_for_conda_based_installations():
...
+        An empty string is returned if a borderline installation is detected
+    """
+    assert os.path.basename(Path(__file__).parents[2]) == "khiops", (
+        "Please fix the `Path.parents` in this method "
+        "so it finds the conda environment directory of this module"
```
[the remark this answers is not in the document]
tramora: Fixed

khiops/core/internals/runner.py
```diff
@@ -189,7 +238,20 @@ def _check_conda_env_bin_dir(conda_env_bin_dir):
 
 
 def _infer_khiops_installation_method(trace=False):
-    """Return the Khiops installation method"""
+    """Return the Khiops installation method
```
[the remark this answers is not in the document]
tramora: Fixed

khiops/core/internals/runner.py
```diff
@@ -412,46 +497,66 @@ def _build_status_message(self):
...
+            A 3-tuple containing in this order :
             - The status message
-            - A list of warning messages
+            - A list of error messages (str)
+            - A list of warning messages (str)
         """
-        # Capture the status of the the samples dir
-        warning_list = []
+        # Capture the status of the samples dir
+        warnings_list = []
```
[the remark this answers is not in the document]
tramora: search replace performed for both `warning_list` and `error_list`

khiops/core/internals/runner.py
```diff
@@ -412,46 +497,66 @@ def _build_status_message(self):
...
-        print("Khiops Python library status OK", end="")
-        if warning_list:
-            print(", with warnings:")
-            for warning in warning_list:
-                print(f"warning: {warning.message}")
-        else:
-            print("")
-        return 0
+        if errors_list or warnings_list:
+            print("Installation issues were detected:\n")
```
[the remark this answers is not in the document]
tramora: Fixed

khiops/core/internals/runner.py
```diff
@@ -972,21 +1077,203 @@ def _initialize_khiops_version(self):
 
         self._khiops_version = KhiopsVersion(khiops_version_str)
 
-        # Warn if the khiops version is too far from the Khiops Python library version
+        # Warn if the khiops version does not match the Khiops Python library version
+        # Currently the check is very strict
+        # (major.minor.patch must be the same)
```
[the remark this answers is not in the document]
tramora: I used to put the same message but @popescu-v asked me to remove it.
Comment modified

khiops/core/internals/runner.py
```diff
@@ -972,21 +1077,203 @@ def _initialize_khiops_version(self):
...
-                "we recommend you to update to the latest Khiops Python "
-                "library version. See https://khiops.org for more information.",
+                "we recommend to update either Khiops binaries or "
+                "the Khiops Python library. "
+                "See https://khiops.org for more information.",
                 stacklevel=3,
             )
 
+    def _detect_library_installation_incompatibilities(self, library_root_dir):
+        """Detect known incompatible installations of this python library
```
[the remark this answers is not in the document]
tramora: Fixed

khiops/core/internals/runner.py
```diff
@@ -972,21 +1077,203 @@ def _initialize_khiops_version(self):
...
+                - a list of error messages
+                - a list of warning messages
+        """
+
+        errors_list = []
+        warnings_list = []
+
+        installation_method = _infer_khiops_installation_method()
+        if installation_method == "conda":
+            # activated 'conda' installation
```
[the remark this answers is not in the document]
tramora: changed

khiops/core/internals/runner.py
```diff
@@ -972,21 +1077,203 @@ def _initialize_khiops_version(self):
...
+                warning = (
+                    "Khiops Python library "
+                    "was not installed with 'conda' or 'pip' "
+                    f"but with '{current_library_installer}' installer. "
+                    "This will probably lead to unexpected errors. "
+                    "Go to https://khiops.org for instructions to re-install it.\n"
+                )
+                warnings_list.append(warning)
+
+            # let's consider only the 'binary+pip' and 'conda-based' installations here
```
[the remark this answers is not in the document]
tramora: changed

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -10,7 +10,12 @@ ARG KHIOPS_REVISION
 RUN true \
   # Install git (for khiops-python version calculation) and pip \
   && apt-get -y update \
-  && apt-get -y --no-install-recommends install git python3-pip zip pandoc wget ruby-dev \
+  && apt-get -y --no-install-recommends install git python3-pip \
+  # On Debian/Ubuntu systems, python3-venv must be installed to provide ensurepip.
+  # It is still possible to create venvs using the --without-pip flag but any subsequent
+  # module installation with pip would be impossible
+  python3.10-venv \
+  zip pandoc wget ruby-dev \
```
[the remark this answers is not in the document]
folmos-at-orange: Ok, I thought they were in this PR.

You may resolve after reading this.

khiops/core/internals/runner.py
```diff
@@ -233,6 +295,29 @@ def _check_executable(bin_path):
         )
 
 
+def _get_current_library_installer():
```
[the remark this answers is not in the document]
folmos-at-orange: After retesting it seems that my environment wasn't ok. 

You may close when reading this.

khiops/core/internals/runner.py
```diff
@@ -412,46 +497,66 @@ def _build_status_message(self):
...
-        else:
-            print("")
-        return 0
+        if errors_list or warnings_list:
+            print("Installation issues were detected:\n")
+            print("---\n")
+
+        # Print the errors (if any)
+        if errors_list:
+            print("Errors to be fixed:")
```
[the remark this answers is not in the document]
folmos-at-orange: Ok

folmos-at-orange (approved):

khiops/core/internals/runner.py
```diff
@@ -412,46 +497,66 @@ def _build_status_message(self):
...
-        try:
-            status_msg, warning_list = self._build_status_message()
-        except (KhiopsEnvironmentError, KhiopsRuntimeError) as error:
-            print(f"Khiops Python library status KO: {error}")
-            return 1
+
+        status_msg, errors_list, warnings_list = self._build_status_message()
 
         # Print status details
         print(status_msg, end="")
```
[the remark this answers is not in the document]
tramora: Line removed

khiops/core/internals/runner.py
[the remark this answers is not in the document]

khiops/core/internals/runner.py
popescu-v: OK, then add specific warning to the user, as discussed.

popescu-v (changes requested):
Add specific user warning if installation type is "unknown" because package metadata is missing (see https://github.com/KhiopsML/khiops-python/pull/402/files#r2284574608 ).

khiops/core/internals/runner.py
```diff
@@ -189,7 +241,19 @@ def _check_conda_env_bin_dir(conda_env_bin_dir):
...
+    """Returns the Khiops installation method
+
+    Definitions :
+    - 'conda' environment contains binaries, shared libraries and python libraries
+    - 'conda-based' environment is similar to 'conda' except that
+       it was not activated previously nor during the execution
+       and thus the CONDA_PREFIX environment variable is undefined
+       and the path to the `bin` directory inside the conda environment is not in PATH
+    - 'binary+pip' installs the binaries and the shared libraries system-wide
+      but will keep the python libraries
```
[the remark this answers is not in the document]
tramora: docstring modified (for the `pip install --user` usage)

khiops/core/internals/runner.py
```diff
@@ -972,21 +1078,203 @@ def _initialize_khiops_version(self):
...
+                            f"'{sys.base_prefix}' "
+                            "should have been used. "
+                            "Go to https://khiops.org for instructions "
+                            "to re-install it "
+                            "(preferably in a virtual environment).\n"
+                        )
+                        error_list.append(error)
+                    # fetch the 'User site' site-packages
+                    # the path is adapted for each OS (Windows, MacOS, Linux)
+                    user_site_packages = site.getusersitepackages()
```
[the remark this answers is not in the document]
tramora: variable name was refactored

tramora:
> Add specific user warning if installation type is "unknown" because package metadata is missing (see https://github.com/KhiopsML/khiops-python/pull/402/files#r2284574608 ).

Warning added

tramora:
> Technically, LGTM. See however a few minor comments (nice to have fixed IMHO). Also update commit message line: "- Detect installation incompatibilities in a system-wide pip installation" to "- Detect installation incompatibilities in a system-wide or user-site pip installation"

Commit message updated

khiops/core/internals/runner.py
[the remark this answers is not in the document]
folmos-at-orange: So we keep it with the `may be loosened`. I think it doesn't hurt.

PR #421, opened 2025-06-23
Add support for the new multi-table schema spec in the s…

popescu-v:
- support the new syntax for X
- deprecate the old syntax: warn the user and transform it to the new syntax.

---

### DIFF
--- a/khiops/sklearn/dataset.py
+++ b/khiops/sklearn/dataset.py
@@ -176,7 +176,7 @@ def _check_multitable_spec(ds_spec):
             )
 
 
-def _table_name_of_path(table_path):
+def table_name_of_path(table_path):
     return table_path.split("/")[-1]
 
 
@@ -387,7 +387,6 @@ class Dataset:
         # Initialize members
         self.main_table = None
         self.additional_data_tables = None
-        self.relations = None
         self.categorical_target = categorical_target
         self.target_column = None
         self.target_column_id = None
@@ -437,7 +436,8 @@ class Dataset:
         # Index the tables by name
         self._tables_by_name = {
             table.name: table
-            for table in [self.main_table] + self.additional_data_tables
+            for table in [self.main_table]
+            + [table for _, table, _ in self.additional_data_tables]
         }
 
         # Post-conditions
@@ -513,32 +513,21 @@ class Dataset:
                 key=main_table_key,
             )
             self.additional_data_tables = []
-            self.relations = []
             if "additional_data_tables" in X:
                 for table_path, table_spec in X["additional_data_tables"].items():
                     table_source, table_key = table_spec[:2]
-                    table_name = _table_name_of_path(table_path)
+                    table_name = table_name_of_path(table_path)
                     table = PandasTable(
                         table_name,
                         table_source,
-                        data_path=table_path,
                         key=table_key,
                     )
-                    self.additional_data_tables.append(table)
                     is_one_to_one_relation = False
                     if len(table_spec) == 3 and table_spec[2] is True:
                         is_one_to_one_relation = True
 
-                    # Set relation parent: if no "/" in path, main_table is the parent
-                    if not "/" in table_path:
-                        parent_table_name = self.main_table.name
-                    else:
-                        table_path_fragments = table_path.split("/")
-                        parent_table_name = _table_name_of_path(
-                            "/".join(table_path_fragments[:-1])
-                        )
-                    self.relations.append(
-                        (parent_table_name, table_name, is_one_to_one_relation)
+                    self.additional_data_tables.append(
+                        (table_path, table, is_one_to_one_relation)
                     )
         # Initialize a sparse dataset (monotable)
         elif isinstance(main_table_source, sp.spmatrix):
@@ -548,7 +537,6 @@ class Dataset:
                 key=main_table_key,
             )
             self.additional_data_tables = []
-            self.relations = []
         # Initialize a numpyarray dataset (monotable)
         elif hasattr(main_table_source, "__array__"):
             self.main_table = NumpyTable(
@@ -561,7 +549,6 @@ class Dataset:
                     "with pandas dataframe source tables"
                 )
             self.additional_data_tables = []
-            self.relations = []
         else:
             raise TypeError(
                 type_error_message(
@@ -680,11 +667,12 @@ class Dataset:
         ds_spec = {}
         ds_spec["main_table"] = (self.main_table.data_source, self.main_table.key)
         ds_spec["additional_data_tables"] = {}
-        for table in self.additional_data_tables:
-            assert table.data_path is not None
-            ds_spec["additional_data_tables"][table.data_path] = (
+        for table_path, table, is_one_to_one_relation in self.additional_data_tables:
+            assert table_path is not None
+            ds_spec["additional_data_tables"][table_path] = (
                 table.data_source,
                 table.key,
+                is_one_to_one_relation,
             )
 
         return ds_spec
@@ -748,31 +736,32 @@ class Dataset:
         # Note: In general 'name' and 'object_type' fields of Variable can be different
         if self.additional_data_tables:
             main_dictionary.root = True
-            table_names = [table.name for table in self.additional_data_tables]
-            tables_to_visit = [self.main_table.name]
-            while tables_to_visit:
-                current_table = tables_to_visit.pop(0)
-                for relation in self.relations:
-                    parent_table, child_table, is_one_to_one_relation = relation
-                    if parent_table == current_table:
-                        tables_to_visit.append(child_table)
-                        parent_table_name = parent_table
-                        index_table = table_names.index(child_table)
-                        table = self.additional_data_tables[index_table]
-                        parent_table_dictionary = dictionary_domain.get_dictionary(
-                            parent_table_name
-                        )
-                        dictionary = table.create_khiops_dictionary()
-                        dictionary_domain.add_dictionary(dictionary)
-                        table_variable = kh.Variable()
-                        if is_one_to_one_relation:
-                            table_variable.type = "Entity"
-                        else:
-                            table_variable.type = "Table"
-                        table_variable.name = table.name
-                        table_variable.object_type = table.name
-                        parent_table_dictionary.add_variable(table_variable)
+            for (
+                table_path,
+                table,
+                is_one_to_one_relation,
+            ) in self.additional_data_tables:
+                if not "/" in table_path:
+                    parent_table_name = self.main_table.name
+                else:
+                    table_path_fragments = table_path.split("/")
+                    parent_table_name = table_name_of_path(
+                        "/".join(table_path_fragments[:-1])
+                    )
+                parent_table_dictionary = dictionary_domain.get_dictionary(
+                    parent_table_name
+                )
 
+                dictionary = table.create_khiops_dictionary()
+                dictionary_domain.add_dictionary(dictionary)
+                table_variable = kh.Variable()
+                if is_one_to_one_relation:
+                    table_variable.type = "Entity"
+                else:
+                    table_variable.type = "Table"
+                table_variable.name = table.name
+                table_variable.object_type = table.name
+                parent_table_dictionary.add_variable(table_variable)
         return dictionary_domain
 
     def create_table_files_for_khiops(self, output_dir, sort=True):
@@ -811,9 +800,9 @@ class Dataset:
 
         # Create a copy of each secondary table
         secondary_table_paths = {}
-        for table in self.additional_data_tables:
-            assert table.data_path is not None
-            secondary_table_paths[table.data_path] = table.create_table_file_for_khiops(
+        for table_path, table, _ in self.additional_data_tables:
+            assert table_path is not None
+            secondary_table_paths[table_path] = table.create_table_file_for_khiops(
                 output_dir, sort=sort
             )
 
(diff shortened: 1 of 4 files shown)


khiops/sklearn/estimators.py
```diff
@@ -558,33 +558,26 @@ def _transform_deploy_model(
...
-
-            additional_data_tables[data_path] = secondary_table_paths[
-                initial_dictionary_name
-            ]
 
         # Set output path files
         output_dir = self._get_output_dir(computation_dir)
         log_file_path = fs.get_child_path(output_dir, "khiops.log")
         output_data_table_path = fs.get_child_path(output_dir, transformed_file_name)
+        self.verbose = True
```
folmos-at-orange, line 572: This line makes the sklearn tests fail (predict and other methods should not change `___dict__`)

folmos-at-orange (changes requested):

khiops/sklearn/estimators.py
```diff
@@ -1170,13 +1161,13 @@ def _transform_create_deployment_dataset(self, ds, _):
         )
 
         # Create the dataset with the keys table as the main one
-        deploy_dataset_spec["tables"][keys_table_name] = (
+        deploy_dataset_spec["main_table"] = (
```
folmos-at-orange, line 1164: I think here is the problem. This code is creating a secondary table with name "main_table". Before it created a table called "keys_main_table". So what is happening is that the files for the keys are not being created.

khiops/sklearn/estimators.py
```diff
@@ -558,33 +558,26 @@ def _transform_deploy_model(
...
                 f"Dictionary '{dictionary.name}' "
                 f"does not have prefix '{self._khiops_model_prefix}'"
             )
-            initial_dictionary_name = dictionary.name.replace(
-                self._khiops_model_prefix, "", 1
-            )
-
-            additional_data_tables[data_path] = secondary_table_paths[
-                initial_dictionary_name
-            ]
```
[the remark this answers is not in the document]
popescu-v: Not anymore AFAIU.

khiops/core/helpers.py
```diff
@@ -34,6 +37,9 @@ def _build_multi_table_dictionary_domain(
         Name for the new root dictionary
     secondary_table_variable_name : str
         Name, in the root dictionary, for the "table" variable of the secondary table.
+    update_secondary_table_name : bool, default `False`
+        If set, then update the secondary table name according to the secondary
+        table variable name. If not set, keep original table name.
```
folmos-at-orange, line 42: Change to ``` If ``True`` ``` to be coherent with the rest of the doc.

khiops/sklearn/dataset.py
```diff
@@ -48,259 +48,128 @@ def check_dataset_spec(ds_spec):
...
 
 
-def _check_table_key(table_name, key):
+def _check_table_key(table_path, key):
     if key is not None:
-        if not is_list_like(key) and not isinstance(key, str):
+        if not is_list_like(key):
             raise TypeError(
-                type_error_message(f"'{table_name}' table's key", key, str, Sequence)
+                type_error_message(f"'{table_path}' table's key", key, Sequence)
```
folmos-at-orange, line 111: I think we should accept `str` as before. It lightens the notation in the most common case: when there is a single key

khiops/sklearn/dataset.py
```diff
@@ -48,259 +48,128 @@ def check_dataset_spec(ds_spec):
...
-    for i, relation in enumerate(relations_spec, 1):
-        parent_table, child_table = relation[:2]
-        if parent_table == child_table:
+        table_key = table_spec[1]
+        if table_key is None:
             raise ValueError(
-                f"Relation #{i}'s tables are equal: ({parent_table}, {child_table}). "
-                "They must be different."
+                f"Key of secondary table at path '{table_path}' is 'None': "
+                "table keys must be specified in multi-table datasets"
```
folmos-at-orange, line 168: `secondary table` -> `additional data table` so it is coherent with the new notation.

khiops/sklearn/dataset.py
```diff
@@ -548,22 +417,19 @@ def _check_input_sequence(self, X, key=None):
         # Check the key for the main_table (it is the same for the others)
         _check_table_key("main_table", key)
 
+    def _table_name_of_path(self, table_path):
+        # TODO: Add >= 128-character truncation and indexing scheme
+        return table_path.split("/")[-1]
```
folmos-at-orange: Make an issue for this and remove this `TODO`

khiops/sklearn/dataset.py
```diff
@@ -573,54 +439,62 @@ def _init_tables_from_mapping(self, X):
                 key=main_table_key,
             )
             self.secondary_tables = []
```
folmos-at-orange, line 501: I think we can rename this `additional_data_tables` to be coherent.

khiops/sklearn/dataset.py
```diff
@@ -573,54 +439,62 @@ def _init_tables_from_mapping(self, X):
...
-            for table_name, (table_source, table_key) in X["tables"].items():
-                if table_name != main_table_name:
-                    self.secondary_tables.append(
-                        PandasTable(table_name, table_source, key=table_key)
+            self.relations = []
```
folmos-at-orange, line 502: Why do we need still need `self.relations` internally ? Should'nt we store a similar `additional_data_tables` field with `data-path -> table` dictionary ?

khiops/sklearn/dataset.py
```diff
@@ -977,13 +841,15 @@ class PandasTable(DatasetTable):
...
+        Data path of the table. Unset for main tables.
     key : list of str, optional
         The names of the columns composing the key.
     """
 
-    def __init__(self, name, dataframe, key=None):
+    def __init__(self, name, dataframe, data_path=None, key=None):
         # Call the parent method
-        super().__init__(name, key=key)
+        super().__init__(name=name, key=key)
```
folmos-at-orange, line 912: This is related to the comment above: The table classes should not know nothing about the data paths, it is the `Dataset` that should have the relation information.

This can be solved by storing an `self.additional_data_tables` dictionary in the `Dataset` class storing the `data-path -> Table` relations.

folmos-at-orange (changes requested):

khiops/sklearn/dataset.py
popescu-v: No, because we need to observe the specification that is listed in the issue, which only mandates the usage of lists and, moreover, ensures that the type of the key is consistent.

khiops/sklearn/dataset.py
popescu-v: Yes, this is a sound argument IMHO, as it would allow us to maintain the Dataset / table encapsulation.

khiops/core/helpers.py
popescu-v: Done.

khiops/sklearn/dataset.py
popescu-v: Done.

khiops/sklearn/dataset.py
folmos-at-orange: I personally missed this detail in the spec, I could have raised an objection. I think it doesn't cost nothing to add this syntax sugar, and will render many examples simpler, as most of the time there is only one key.

popescu-v: Could you file a _separate_ issue for amending the spec in this respect? This would be a basis for a discussion in this respect IMHO.

khiops/sklearn/dataset.py
popescu-v: Done.

khiops/sklearn/dataset.py
folmos-at-orange: Done (https://github.com/KhiopsML/khiops-python/issues/431). You may resolve when you read this.

popescu-v: OK, noted.

khiops/sklearn/dataset.py
```diff
@@ -48,258 +54,173 @@ def check_dataset_spec(ds_spec):
...
-    elif isinstance(parent_table_key, str) and is_list_like(child_table_key):
-        error_found = parent_table_key not in child_table_key
-    elif is_list_like(parent_table_key) and is_list_like(child_table_key):
-        error_found = not set(parent_table_key).issubset(child_table_key)
-    elif is_list_like(parent_table_key) and isinstance(child_table_key, str):
-        error_found = (
-            len(parent_table_key) != 1 or child_table_key not in parent_table_key
-        )
+def table_name_of_path(table_path):
+    return table_path.split("/")[-1]
```
folmos-at-orange, line 181: I think this small function should be defactored for better locality.

It is used twice below, and once by splitting an rejoining the path string.

khiops/sklearn/dataset.py
```diff
@@ -552,18 +474,27 @@ def _init_tables_from_mapping(self, X):
         """Initializes the table spec from a dict-like 'X'"""
         assert is_dict_like(X), "'X' must be dict-like"
 
+        # Detect if deprecated mapping specification syntax is used;
+        # if so, issue deprecation warning and transform it to the new syntax
+        if "tables" in X.keys():
+            warnings.warn(
+                deprecation_message(
```
folmos-at-orange, line 481: just for good measure i'd add `or (isinstance(X.get("main_table"), str)`.

khiops/sklearn/dataset.py
```diff
@@ -48,258 +54,173 @@ def check_dataset_spec(ds_spec):
...
+                        and right in table_path
+                    ):
+                        table_path.insert(0, left)
+            table_path = "/".join(table_path)
+            if is_entity:
+                table_data = (table_df, table_key, is_entity)
+            else:
+                table_data = (table_df, table_key)
+            new_X["additional_data_tables"][table_path] = table_data
+    return new_X
```
folmos-at-orange, line 223: `X` -> `ds_spec`
`new_X` -> `new_ds_spec`.

Also I'd add a few more comments for the `if` cases at least.

khiops/sklearn/dataset.py
```diff
@@ -48,258 +54,173 @@ def check_dataset_spec(ds_spec):
...
+            # Cycle twice on the relations to get all transitive relation, like:
+            # - current table name N
+            # - main table name N1
+            # - and relations: (N1, N2), (N2, N3), (N3, N)
+            # the data-path must be N2/N3/N
+            # If no "relations" key exists, then one has a star schema and
+            # the data-paths are the names of the secondary tables themselves
+            # (with respect to the main table)
+            if "relations" in X:
+                for relation in islice(cycle(X["relations"]), 2 * len(X["relations"])):
```
folmos-at-orange, line 207: I don't think `itertools` is needed here. Simply do
`for relation in X["relations"] * 2`

khiops/sklearn/dataset.py
```diff
@@ -48,258 +54,173 @@ def check_dataset_spec(ds_spec):
...
+            is_entity = False
+
+            # Cycle twice on the relations to get all transitive relation, like:
+            # - current table name N
+            # - main table name N1
+            # - and relations: (N1, N2), (N2, N3), (N3, N)
+            # the data-path must be N2/N3/N
+            # If no "relations" key exists, then one has a star schema and
+            # the data-paths are the names of the secondary tables themselves
+            # (with respect to the main table)
```
folmos-at-orange, line 205: I'm not sure that cycling twice it suffices to get all transitive relations. What about a mega table path `a/b/c/d/e/f/g` does this work ?

folmos-at-orange (changes requested):

khiops/sklearn/dataset.py
popescu-v: Ok.

khiops/sklearn/dataset.py
popescu-v: Why not `and`, because these are both parts of the old spec?

khiops/samples/samples_sklearn.py
```diff
@@ -245,7 +238,7 @@ def khiops_classifier_multitable_snowflake():
     X_train, X_test, y_train, y_test = train_test_split_dataset(X, y)
 
     # Train the classifier (by default it creates 1000 multi-table features)
-    khc = KhiopsClassifier(n_trees=0)
+    khc = KhiopsClassifier(n_trees=0, verbose=True)
```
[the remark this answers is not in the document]
popescu-v: Done: removed the verbose.

khiops/sklearn/dataset.py
popescu-v: On second thought, I would keep it, as it informs on the intention behind this manipulation and, should we change the way the name is computed, there is one single, clearly identified place to do this.

khiops/sklearn/dataset.py
popescu-v: I would change the `if` to:
```python
if "tables" in X.keys() and isinstance(X.get("main_table"), str):
```

khiops/sklearn/dataset.py
popescu-v: Indeed, the number of cyclings should depend on the maximum path length in the relations graphs. So, we need to compute that first. If this doesn't fit the current PR, I plan on adding an issue in this respect. For the current PR, we could set-up a big-enough number of repetitions to account for most currently expected cases (as this is a deprecation path).

khiops/sklearn/dataset.py
popescu-v: Ok, done.

khiops/sklearn/dataset.py
popescu-v: Done for the renamings.

khiops/sklearn/dataset.py
popescu-v: Done.

khiops/sklearn/dataset.py
folmos-at-orange: Ok, but change the weird call below.

khiops/sklearn/dataset.py
folmos-at-orange: Ok

khiops/sklearn/dataset.py
folmos-at-orange: Then cycle 3 times just for good measure.

folmos-at-orange: Also comment that this is an heuristic, the real algorithm should traverse the relations graph.

PR #441, opened 2025-07-29
Migrate to pyproject toml

### LINKED ISSUE #411: Migrate from setup.py to pyproject.toml
#### Description
Using a recent `pip` version, when installing the `khiops` the following message appears:
  > DEPRECATION: Building 'khiops' using the legacy setup.py bdist_wheel mechanism, which will be
  > removed in a future version. pip 25.3 will enforce this behaviour change. A possible replacement
  > is to use the standardized build interface by setting the `--use-pep517` option, (possibly
  > combined with `--no-build-isolation`), or adding a `pyproject.toml` file to the source tree of
  > 'khiops'.  Discussion can be found at https://github.com/pypa/pip/issues/6334

So the `setup.py` will eventually be removed, and the only solution is to migrate the `setup.py` to a `pyproject.toml`. If we do nothing, the `pip` installation method will fail eventually.

#### Questions/Ideas
- Note: A way to build with a `pyproject.toml` file is to use the [`build` package](https://pypi.org/project/build/). Once installed our current build method is equivalent to `python -m build . sdist`

popescu-v:
closes #411 

---

  - [x] Check the docs build **without** warning: see the log of the API Docs workflow
  - [x] Check that your changes render well in HTML: download the API Docs artifact and open `index.html`


.github/workflows/tests.yml
```diff
@@ -105,9 +105,13 @@ jobs:
           do
             # Since Python 3.13, setuptools is not installed automatically anymore
             $CONDA install -y -n "$CONDA_ENV" setuptools
-            $CONDA run --no-capture-output -n "$CONDA_ENV" python setup.py egg_info
-            $CONDA install -y -n "$CONDA_ENV" `grep -v "^\[" khiops.egg-info/requires.txt`
-            rm -rf khiops.egg-info
+
+            # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+            $CONDA install -y -n "$CONDA_ENV" tomli
```
tramora, line 110: Does it mean that under python >= 3.12 this command will install nothing as the tomli lib is already there ?

.github/workflows/tests.yml
```diff
@@ -105,9 +105,13 @@ jobs:
...
+            $CONDA run --no-capture-output -n "$CONDA_ENV" python -c \
+              "import tomli; from functools import reduce; fp = open('pyproject.toml', 'rb'); p = tomli.load(fp); fp.close(); print(' '.join(reduce(lambda x, y: x + y, list(p['project']['optional-dependencies'].values()), p['project']['dependencies'])))" > requires.txt
+            $CONDA install -y -n "$CONDA_ENV" `cat requires.txt`
```
tramora, line 113: I am a bit worried about the complexiification (a not straightforward snippet) to install the dependencies.

.github/workflows/tests.yml
```diff
@@ -252,23 +256,22 @@ jobs:
...
         shell: pwsh
         run: |
           # The following git command is required,
           # as the Git repository is in a directory the current user does not own,
           # Python versioneer fails to compute the current version correctly otherwise
           git config --global --add safe.directory $(Resolve-Path '.' | % {$_.toString()})
-          python setup.py egg_info
+          python -m pip install setuptools
+          python -c `
+            "import tomllib; from functools import reduce; fp = open('pyproject.toml', 'rb'); p = tomllib.load(fp); fp.close(); print('\n'.join(reduce(lambda x, y: x + y, list(p['project']['optional-dependencies'].values()), p['project']['dependencies'])))" > requires.txt
```
tramora, line 269: Duplicate snippet to extract the dependencies. I'm wondering if it is not possible to factorize it

.github/workflows/tests.yml
```diff
@@ -344,9 +347,12 @@ jobs:
           # as the Git repository is in a directory the current user does not own,
           # Python versioneer fails to compute the current version correctly otherwise
           git config --global --add safe.directory $(realpath .)
-          python setup.py egg_info
-          pip install `grep -v "^\[" khiops.egg-info/requires.txt`
-          rm -rf khiops.egg-info
+          # Install tomli for Python < 3.11
+          pip install tomli
+          python -c \
+            "import tomli; from functools import reduce; fp = open('pyproject.toml', 'rb'); p = tomli.load(fp); fp.close(); print(' '.join(reduce(lambda x, y: x + y, list(p['project']['optional-dependencies'].values()), p['project']['dependencies'])))" > requires.txt
```
tramora, line 353: Snippet duplication

.github/workflows/tests.yml
tramora: I understand now it's because of a missing wish-to-have `--only-deps` flag to `pip` (https://github.com/pypa/pip/issues/11440)

.github/workflows/tests.yml
popescu-v: Indeed, it could be factored out (see also comment above).

.github/workflows/tests.yml
tramora: OK

tramora (approved):

tramora (approved):
Looks good

PR #462, opened 2025-08-25
Add n_text_features and type_text features to Khiops sklearn supervised estimators

### LINKED ISSUE #39: Support text types in sklearn predictors
#### Description
Khiops 11 supports `Text` columns which have a specialized AutoML treatment as oppossed to normal strings(`Categorical`). Sklearn predictors should also support this type.

#### Questions/Ideas
- Does pandas and/or numpy have a specialized `Text` type?
- Option 1: Implement it as a `Dataset` property
  - Add to the table specification tuple should have an optional field `text_columns` with the names of the text fields **or**
  - Add another field to the spec `table_text_columns` indexed by the table name and whose values are the names of the text columns (I prefer this one)
  - When creating the dictionary the `Dataset` object will have all the necessary info to add the specified columns as `Text` 
- Option 2: Implement it as a `fit` parameter
  - As above add `table_text_columns` but as a `fit` optional parameter
    - I works but the fact that a column is a `Text` is part of the description of the dataset
    - This parameter should be feeded to the dictionary creation routine
- Expose Dataset API it will have two init patters:
  - Big Constructor 
  - Builder pattern
- The big constructor uses the builder 
- And the `dict` interface should be maintained

popescu-v:
closes #39 

---


khiops/sklearn/dataset.py
```diff
@@ -236,18 +236,25 @@ def _upgrade_mapping_spec(ds_spec):
...
+def get_khiops_type(numpy_type, max_type_size=None):
     """Translates a numpy dtype to a Khiops dictionary type
 
     Parameters
     ----------
-    numpy_type : `numpy.dtype`:
+    numpy_type : `numpy.dtype`
         Numpy type of the column
+    max_type_size : `int`, optional
+        Maximum length of the entries of the column whose type is ``numpy_type``.
```
folmos-at-orange, line 247: I'd rename `max_type_size` to `categorical_str_max_size` or something similar as it is used only in that context.

khiops/sklearn/dataset.py
```diff
@@ -257,6 +264,18 @@ def get_khiops_type(numpy_type):
     # float<x>, int<x>, uint<x> types
     elif "int" in lower_numpy_type or "float" in lower_numpy_type:
         khiops_type = "Numerical"
+    elif lower_numpy_type == "string":
+        if max_type_size is not None:
+            if not isinstance(max_type_size, (int, np.int64)):
+                raise TypeError(
+                    type_error_message("max_type_size", max_type_size, int, np.int64)
+                )
```
folmos-at-orange, line 272: Type checks must be done at the beginning of the method to render the logic more readable.

khiops/sklearn/dataset.py
```diff
@@ -956,10 +975,14 @@ def __init__(self, name, dataframe, key=None):
...
         # Initialize Khiops types
-        self.khiops_types = {
-            column_id: get_khiops_type(self.data_source.dtypes[column_id])
-            for column_id in self.column_ids
-        }
+        self.khiops_types = {}
+        for column_id in self.column_ids:
+            column = self.data_source[column_id]
+            column_numpy_type = column.dtype
+            column_max_size = column.astype("str").str.len().max()
```
folmos-at-orange, line 982: This calculation should be done only for `StringDType`. This avoids the cast and computation of the max on unrelated columns.

```python
column_max_size = None
if isinstance(column_numpy_type, StringDtype):
    column_max_size = column.str.len().max()
```

khiops/samples/samples_sklearn.py
```diff
@@ -24,6 +24,75 @@
...
 
+def khiops_classifier_text():
+    """Train a `.KhiopsClassifier` on a monotable dataframe with textual data"""
+    # Imports
+    import os
+    import pandas as pd
+    from khiops import core as kh
+    from khiops.sklearn import KhiopsClassifier
+    from sklearn import metrics
+    from sklearn.model_selection import train_test_split
```
folmos-at-orange, line 35: I'd move this method below the multiclass examples.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 ### Added
 - (`core`) Dictionary API support for dictionary, variable and variable block
   comments, and dictionary and variable block internal comments.
+- (`sklearn`) "Text" Khiops type support at the estimator level.
```
folmos-at-orange, line 14: `"Text"` -> ``` `Text` ```

folmos-at-orange (changes requested):

khiops/sklearn/estimators.py
```diff
@@ -1548,6 +1568,8 @@ def __init__(
         self,
         n_features=100,
         n_trees=10,
+        n_text_features=10000,
```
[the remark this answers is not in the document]
popescu-v: Good point. Indeed, each single-column with such text data would be internally converted by Khiops into a set of potentially numerous text variables. But I'm not sure of the performance penalty this entails (if any). Setting `n_text_features=0` just tells Khiops to proceed in the "standard" way, where each column is either Numeric or Categorical.

khiops/sklearn/dataset.py
popescu-v: Yes, but this is only relevant for the "string" DType.

doc/samples/samples_sklearn.rst
```diff
@@ -95,6 +95,72 @@ Samples
...
+
+    # Split the dataset into:
+    # - the X feature table
+    # - the y target vector ("negativereason" column)
+    X_train = data_train_df.drop("negativereason", axis=1)
+    X_test = data_test_df.drop("negativereason", axis=1)
+    y_train = data_train_df["negativereason"]
+    y_test = data_test_df["negativereason"]
+
+    # Set Pandas StringDType on the "text" column
```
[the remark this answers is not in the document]
popescu-v: > On the user point of view, requiring this is a burden. If he forgets to do this, does it fail ?

If the user forgets to force this conversion, then the column would either:
- be of a different type (e.g. `object`), in which case it will be assigned the "Categorical" Khiops type;
- or just happen to be of the String dtype, in which case the heuristic would apply.

Hence, there is no failure, but this behavior needs to be thoroughly documented.

khiops/sklearn/dataset.py
```diff
@@ -257,6 +264,18 @@ def get_khiops_type(numpy_type):
     # float<x>, int<x>, uint<x> types
     elif "int" in lower_numpy_type or "float" in lower_numpy_type:
         khiops_type = "Numerical"
+    elif lower_numpy_type == "string":
+        if max_type_size is not None:
+            if not isinstance(max_type_size, (int, np.int64)):
+                raise TypeError(
+                    type_error_message("max_type_size", max_type_size, int, np.int64)
+                )
+            if max_type_size > 100:
```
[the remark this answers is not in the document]
popescu-v: See above. I'm not sure, as it is only used once.

khiops/sklearn/dataset.py
popescu-v: Done.

khiops/sklearn/dataset.py
popescu-v: Done.

CHANGELOG.md
popescu-v: Done.

folmos-at-orange (approved):
I'd merge the commit for the feature with that of the sample, otherwise LGTM.

Issue #482, opened 2025-09-30
KhiopsJSONError raised with coclustering report containing a numerical variable with one part (all elements).


pierrenodet:
#### Description
A KhiopsJSONError is raised when attempting to load a coclustering report with a numerical variable that has only one part (all elements are within the same interval, [min; max]).

#### Questions/Ideas
https://github.com/KhiopsML/khiops-python/blob/c1c4c3502b855fd4f5c7624bc9b853c572999c62/khiops/core/coclustering_results.py#L913

I don't understand why this check for length >=2 is mandatory. I understand that it is done if the first part contains the missing values, because you would need at least a second one with the rest of the samples. However, what if I only have one cluster with all the elements? Maybe the actual check is too restrictive. I can load this coclustering report in the coviz app.

To reproduce, simply add a new variable to the dataset before performing coclustering, which is white noise, and then try loading the coclustering report with khiops-python.

popescu-v:
Closed by PR #484

PR #484, opened 2025-09-30
Fix interval length check in coclustering report parsing

pierrenodet:
_Put your message here_

---

Fix issue #482 

### DIFF
--- a/khiops/core/coclustering_results.py
+++ b/khiops/core/coclustering_results.py
@@ -432,7 +432,7 @@ class CoclusteringReport:
             # Check minimum consistency of input data
             if "cellFrequencies" not in json_data:
                 raise KhiopsJSONError(
-                    "'cellFrequencies' key not found " "but 'cellPartIndexes' found."
+                    "'cellFrequencies' key not found but 'cellPartIndexes' found."
                 )
             json_cell_frequencies = json_data["cellFrequencies"]
             json_cell_part_indexes = json_data["cellPartIndexes"]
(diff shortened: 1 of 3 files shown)


khiops/core/coclustering_results.py
```diff
@@ -945,7 +945,6 @@ def init_partition(self, json_data=None):
 
             # Instance-variable coclustering: initialize inner variables
             if self.is_variable_part:
-
```
popescu-v, line 948: This should be kept, as it's a paragraph (starting with a comment line), and paragraphs are (usually) preceded by an empty line.

popescu-v (changes requested):
A small change (see comment).

khiops/core/coclustering_results.py
```diff
@@ -909,8 +909,6 @@ def init_partition(self, json_data=None):
                 )
             elif not isinstance(json_data["intervals"], list):
                 raise KhiopsJSONError("'intervals' key must be a list")
-            elif len(json_data["intervals"]) < 2:
```
popescu-v, line 912: We should check however that there is at least one interval here (otherwise `self.parts[0]` will fail with `IndexError` below).

popescu-v (changes requested):
Add check that there is at least one interval (see relevant comment).
Squash all commits into one. Provide a more explanatory commit message, like:
```
Fix interval length check in coclustering report parsing

Systematically check that there is at least one interval.

Check that there are at least two intervals only if the first interval contains only missing values.
```

popescu-v (approved):

PR #496, opened 2025-10-20
Test Git tag / package version coherence

popescu-v:
Put this check as the last step of the `test` job and have it conditioned by the workflow being launched on a Git tag, in order to easily factor it out among the 3 potential target release "channels" (GitHub, Test PyPI, and PyPI).

- test with non-matching tags: https://github.com/KhiopsML/khiops-python/actions/runs/18655441897.
- test with matching tags: https://github.com/KhiopsML/khiops-python/actions/runs/18655640342.

---

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -238,6 +238,24 @@ jobs:
           # See the upload-artifact step in the build job for the explanation of this pattern
           name: khiops-conda
           path: ./khiops-conda
+      - name: Check Conda package / Git tag version coherence
+        run: |
+          # Don't exit on first error: print relevant error message
+          set +e
+          # Extract Conda package version from the package artifact
+          PACKAGE_VERSION=$( \
+            unzip -p ./khiops-conda/noarch/*.conda info-*.tar.zst \
+            | tar --zstd -axf - info/index.json -O \
+            | jq -r ".version" \
+          )
+          # Convert pre-release version specification in the Git tag to the
+          # Conda format and check that it matches the Conda package version
+          echo ${{ github.ref_name }} | tr -d '-' | grep -wq $PACKAGE_VERSION
+          if [[ $? -ne 0 ]]
+          then
+            echo "::error::Conda package version $PACKAGE_VERSION does not match Git tag ${{ github.ref_name }}"
+            false
+          fi
       - name: Install Miniforge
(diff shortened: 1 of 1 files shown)


.github/workflows/pip.yml
```diff
@@ -120,6 +123,21 @@ jobs:
...
           # (as given by khiops_env)
           kh-status | grep "MPI command" | grep -vwq "<empty>"
+      - name: Test package / Git tag version coherence
+        shell: bash
+        if: github.ref_type == 'tag'
+        run: |
+          # Don't exit on first error: print relevant error message
+          set +e
+          PACKAGE_VERSION=$(pip show khiops | awk 'BEGIN {FS = ": "} $1 ~ /^Version$/ {print $2}')
+          # Convert pre-release version specification to the Pip format
```
folmos-at-orange: I'd just add the `PACKAGE_VERSION=...` statement to the paragraph below.

popescu-v: Ok & ok. One key difference wrt. the Conda packaging workflow check is that here we check the Pip package metadata directly, rather than the setting in the `khiops/__init__.py` file.

On the other hand, IMHO this discussion shows that we also need to check the Conda package version against the Git tag (in the `conda.yml` workflow, we only check that the Conda package version as specified in the `packaging/conda/meta.yaml` matches the version specified in `khiops/__init__.py` file). I intend to add a commit along these lines in the current PR.

folmos-at-orange: Doesn't the `meta.yml` obtains the version dynamically ?

.github/workflows/conda.yml
```diff
@@ -238,6 +238,26 @@ jobs:
...
+      - name: Check Conda package / Git tag version coherence
+        run: |
+          # Don't exit on first error: print relevant error message
+          set +e
+          # Extract Conda package version from the package artifact, without JQ,
+          # which would require having a Conda environment set-up
+          PACKAGE_VERSION=$( \
+            unzip -p ./khiops-conda/noarch/*.conda info-*.tar.zst \
+            | tar --zstd -axf - info/index.json -O \
+            | awk 'BEGIN {FS = ": "} /"version"/ {print $2}' \
```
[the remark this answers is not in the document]
popescu-v: Done: used `jq`.

PR #512, opened 2025-12-04
Drop discarded sample dataset

popescu-v:


popescu-v:
Full, expensive tests pass: https://github.com/KhiopsML/khiops-python/actions/runs/19937352127.

Issue #382, opened 2025-04-03
Add Automatic "Unhappy Path" Tests in the CI


popescu-v:
#### Description
To make sure "unhappy path" warnings keep behaving as expected, we need to test them automatically and systematically in the CI.

Issue #301 is concerned with the support for these warnings. The current issue is concerned with testing that this support keeps functioning as expected.

PR #516, opened 2025-12-12
Improve messages and behaviour when detecting unhappy installation paths

### LINKED ISSUE #515: Requested message improvements for the unhappy installation paths
#### Description
Peer testers state the current **messages shown during the unhappy installation paths dectection** need improvements.

At least in two cases :
- **version check** : currently the libary shows a generic warning message when the library version triplet (major, minor, patch) does not match the one from khiops core
- **detection of a mix of pip and conda** : the tip of asking to installing in the current conda environment is misleading  

#### New messages proposal

For the version check : 
- if the major version differs, a error message (instead of a warning one) must be raised
- if the minor or patch differ, a warning will be raised as currently

In case of a detection of a mix of pip and conda, the tip should be clearly choosing between the too in disabling the one not used.

tramora:
- in case of major version mismatch, raise immediately an error
- in case of a mix between conda and something else (pip or a cloned source folder), we suggest deactivating the conda env or install the library inside this later
- discourage using system-wide khiops binary and a library inside a conda env

Fixes #515 

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -445,16 +445,29 @@ jobs:
           fi
 
           # Print status
-          python -c \
-            "import sys; import khiops.core as kh; return_code = kh.get_runner().print_status(); sys.exit(return_code)"
-          RETCODE=$?
+          PATTERN=$(python -c \
+            "import sys; import khiops.core as kh; return_code = kh.get_runner().print_status(); sys.exit(return_code)" 2> >(grep -Ei 'with.*?unknown.*?installer') )                    
           # The installation status check will fail
           # because the library was not installed properly (the source code was only cloned)
-          if [ $RETCODE -ne 0 ]; then
+          if [ -n "$PATTERN" ]; then
             echo "::error::Status error: improper setup, as expected: khiops-python has been cloned, not installed from a package"
           fi
 
-          # Run integration tests on Ubuntu and Rocky
+          # Run the library against an incompatible Khiops (with a different major version)
+          # This instance of Khiops is isolated in a dedicated conda environment
+          CONDA="/root/miniforge3/bin/conda"                
+          # Check an error is raised because of the major version mismatch
+          # The khiops-python library from the cloned sources is used here
+          PATTERN=$($CONDA run -n py3_khiops10_conda python -c "import khiops.core as kh; print(kh.get_runner().khiops_version)" 2> >(grep -Ei 'major version.*?does not match'))          
+          if [ -z "$PATTERN" ]; then
+            echo "::error::Status error: khiops-python should fail because of the major version mismatch"          
+            if [[ "${{ matrix.container }}" == "debian13" ]]; then            
+              deactivate                      
+            fi          
+            exit 1;
+          fi
+
+          # Run the remaining integration tests
           python -m unittest -v tests.test_khiops_integrations
 
           # Execute Khiops sample (train and deploy model), which uses MPI
--- a/packaging/docker/khiopspydev/Dockerfile.debian
+++ b/packaging/docker/khiopspydev/Dockerfile.debian
@@ -74,7 +74,12 @@ RUN true \
         $CONDA install -y -n py${version}_conda \
         khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
         khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION}; \
-    done' \
+    done; \
+    # Install Khiops from a different major version in a dedicated conda environment \
+    # The python interpreter version of the base environment is used as no specific version is given \
+    $CONDA create -y -n py3_khiops10_conda; \
+    $CONDA install -y -n py3_khiops10_conda khiops::khiops-core==10.3.1; \
+    ' \
   && true
 
 RUN mkdir -p /scripts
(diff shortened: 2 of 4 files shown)


khiops/core/internals/runner.py
popescu-v: IMHO we can discard the `.major` field in these checks, because if the major fields are different one raises an exception before anyway and never reach these lines. We only reach these lines if the major versions are identical between Core and the Python library.

tramora:
The users may be surprised that the major version mismatch raises an exception and cuts the running flow immediately.
I could be changed to delay this and show then a "clean" installation status report.
The major drawback of this alternative is that the users may never run the installation status check and run tasks against an incompatible version of Khiops. The implementation of this PR seems then the best way to raise this error. It the users complain we could think about.

tramora:
> See the comment. I also think that we should add a test that an exception is raised if the major versions are different between Core and the Python library, in `test_khiops_integrations.py` plus changes in the CI. Since this test is out of the scope of this PR, I would update issue #382 accordingly with a comment.

In my opinion this very border-case of incompatible major version is not worth being implemented into `test_khiops_integrations.py` and if so it would technically more difficult to reach the dedicated conda environment I'm thinking putting a Khiops10 installation.

It seems simplier to me to stick to the setup into the docker dev images and test directly within the CI workflow (before running `test_khiops_integrations.py`)

popescu-v:
> It seems simplier to me to stick to the setup into the docker dev images and test directly within the CI workflow (before running `test_khiops_integrations.py`)

Yes, but we need to execute the Khiops Python code to make sure the exception is raised. Something like launching `kh-status` in the CI and checking the return code `!= 0` and the contents of `stderr` might suffice though. Is this what you have in mind @tramora ? If so, then can you update issue #382 accordingly?

.github/workflows/tests.yml
```diff
@@ -454,7 +454,19 @@ jobs:
             echo "::error::Status error: improper setup, as expected: khiops-python has been cloned, not installed from a package"
           fi
 
-          # Run integration tests on Ubuntu and Rocky
+          # Run the library against an incompatible Khiops (with a different major version)
+          # This instance of Khiops is isolated in a dedicated conda environment
+          CONDA="/root/miniforge3/bin/conda"
+          KHIOPS10_CONDA_ENV=py3_khiops10_conda          
+          # Install the library from the sources via pip in this dedicated conda environment
+          $CONDA run -n $KHIOPS10_CONDA_ENV pip install .
```
popescu-v: Rather than doing `pip install` inside a Conda environment, which is an unhappy path as per our own support policy, I'd just:
1. activate the Conda environment (hence these would be the MODL* binaries in the PATH)
2. `cd` to `khiops-python`, then install its dependencies inside the Conda environment, similarly to what it is done (in the context of Pip environments though) as it is done in the `check-khiops-integration-on-linux` job of the `tests.yml` workflow.

.github/workflows/tests.yml
```diff
@@ -454,7 +454,23 @@ jobs:
...
-          # Run integration tests on Ubuntu and Rocky
+          # Run the library against an incompatible Khiops (with a different major version)
+          # This instance of Khiops is isolated in a dedicated conda environment
+          CONDA="/root/miniforge3/bin/conda"                
+          # Check an error is raised because of the major version mismatch
+          # The khiops-python library from the cloned sources is used here
+          output=$($CONDA run -n py3_khiops10_conda python -c "import khiops.core as kh; print(kh.get_runner().khiops_version)" 2>&1)
+          echo ${output}|grep -Ei 'major version.*?does not match'
+          RETCODE=$?
+          if [ $RETCODE -ne 0 ]; then
```
popescu-v: Unset `OUTPUT` before this line.

.github/workflows/tests.yml
```diff
@@ -454,7 +454,23 @@ jobs:
...
           fi
 
-          # Run integration tests on Ubuntu and Rocky
+          # Run the library against an incompatible Khiops (with a different major version)
+          # This instance of Khiops is isolated in a dedicated conda environment
+          CONDA="/root/miniforge3/bin/conda"                
+          # Check an error is raised because of the major version mismatch
+          # The khiops-python library from the cloned sources is used here
+          output=$($CONDA run -n py3_khiops10_conda python -c "import khiops.core as kh; print(kh.get_runner().khiops_version)" 2>&1)
+          echo ${output}|grep -Ei 'major version.*?does not match'
```
popescu-v: Add spaces around the pipe, for legibility reasons.

popescu-v (changes requested):
Just a few comments left.

.github/workflows/tests.yml
```diff
@@ -454,7 +454,23 @@ jobs:
             echo "::error::Status error: improper setup, as expected: khiops-python has been cloned, not installed from a package"
           fi
 
-          # Run integration tests on Ubuntu and Rocky
+          # Run the library against an incompatible Khiops (with a different major version)
+          # This instance of Khiops is isolated in a dedicated conda environment
+          CONDA="/root/miniforge3/bin/conda"                
+          # Check an error is raised because of the major version mismatch
+          # The khiops-python library from the cloned sources is used here
+          output=$($CONDA run -n py3_khiops10_conda python -c "import khiops.core as kh; print(kh.get_runner().khiops_version)" 2>&1)
```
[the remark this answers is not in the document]
tramora: Refactored

PR #529, opened 2026-01-12
Make all sklearn estimators concurrency-safe

### LINKED ISSUE #524: Temporary directory handling renders sklearn estimators non-thread safe
#### Description
For any sklearn action (`fit`, `predict*` or `transform`), the sklearn estimators create a dedicated temporary directory (`computation_dir`) to store intermediary files.

This should be enough to ensure thread-safety (ex. each thread executing `predict` with the same trained classifier). However, the estimator **currently sets the khiops' runner temporary directory to `computation_dir`**. This implies that if two concurrent threads execute `predict` on the same object (or even in different ones!), there is a data race for the khiops' runner temporary directory.

#### Questions/Ideas
- My idea to use the `computation_dir` as the runner's temporary dir was to have a minor convenience for the `verbose/trace` mode, as all files related to an action would be stored in the same directory. Evidently this wasn't a good call.
- A simple solution is to not modify the khiops runner's temp dir in each action. The files will be stored in various directories, but the `verbose/trace` mode will report each of them. This will also simplify the code.
- To reproduce (with high probability)
```python
import khiops.core as kh
import pandas as pd
from khiops.sklearn import KhiopsClassifier
from concurrent.futures import ThreadPoolExecutor, as_completed


def pred(clf, X):
    return clf.predict(X)


adult_df = pd.read_csv(f"{kh.get_samples_dir()}/Adult/Adult.txt", sep="\t", header=0)

clf = KhiopsClassifier(n_trees=0)
clf.fit(adult_df.drop("class", axis=1), adult_df["class"])

with ThreadPoolExecutor(max_workers=10) as executor:
    futures = {
        executor.submit(pred, clf, adult_df.drop("class", axis=1)): i for i in range(10)
    }

    for future in as_completed(futures):
        print("INDEX:", futures[future])
        print(future.result())

```

tramora:
Fixes #524 

---

### DIFF
--- a/tests/test_sklearn.py
+++ b/tests/test_sklearn.py
@@ -6,18 +6,24 @@
 ######################################################################################
 """Tests parameter transfer between Khiops sklearn and core APIs"""
 import contextlib
+import io
 import os
 import shutil
 import unittest
 import warnings
+from concurrent.futures import as_completed
+from concurrent.futures.thread import ThreadPoolExecutor
 from itertools import product
 
 import numpy as np
+import pandas as pd
 from sklearn.exceptions import NotFittedError
 from sklearn.utils.estimator_checks import check_estimator
 from sklearn.utils.validation import check_is_fitted
 
 import khiops.core as kh
+import khiops.core.internals.filesystems as fs
+from khiops.core.internals.runner import KhiopsLocalRunner
 from khiops.sklearn.estimators import (
     KhiopsClassifier,
     KhiopsCoclustering,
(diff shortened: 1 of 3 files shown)


tests/test_sklearn.py
```diff
@@ -1773,6 +1780,27 @@ def test_sklearn_check_estimator(self):
                     print("Done")
 
 
+def disable_mpi_decorator(func):
```
popescu-v: I'd rename to just `no_mpi` or `disable_mpi`.

PR #554, opened 2026-02-13
Fix the detection of an unhappy installation path in a venv under Windows

### LINKED ISSUE #548: Spurious Warning Occurs in Binary + Pip installations on Windows
#### Description
When `khiops` is installed in Python virtual environments generated by the official VSCode Python plugin on Windows, we get the following spurious warning, even though Khiops is installed correctly:
```
"Khiops Python library installation path '...' does not match the current python environment ''."
"Go to https://khiops.org for instructions to re-install it (preferably in a virtual environment).\n"
```

This is due to the fact that the capitalization of the base directory `base_dir` sometimes differs from the capitalization of `sys.executable`, in terms of the drive letter `X:\` vs `x:\`.

#### Questions/Ideas
- The solution consists in lower-casing `base_dir` and `sys.executable` when comparing them on Windows, because on Windows paths are not case-sensitive.

tramora:
External tools like VSCode can modify the case of the path saved in `sys.executable`

Fixes #548 

The case is indeed reproductible in saving the value of `sys.executable` and altering its case.

Implementing an automatic test would required extracting a function returning `sys.executable` and mocking it during the test.
Possible but but a bit overkill.

---

### DIFF
--- a/khiops/core/internals/runner.py
+++ b/khiops/core/internals/runner.py
@@ -542,14 +542,14 @@ class KhiopsRunner(ABC):
         assert (
             os.path.basename(Path(__file__).parents[2]) == "khiops"
         ), "Please fix the `Path.parents` in this method "
-        library_root_dir = Path(__file__).parents[2]
+        library_root_dir_path = Path(__file__).parents[2]
 
         status_msg = "Khiops Python library settings\n"
         status_msg += f"version             : {khiops.__version__}\n"
         status_msg += f"runner class        : {self.__class__.__name__}\n"
         status_msg += f"root temp dir       : {self.root_temp_dir}\n"
         status_msg += f"sample datasets dir : {samples_dir_path}\n"
-        status_msg += f"library root dir    : {library_root_dir}\n"
+        status_msg += f"library root dir    : {library_root_dir_path}\n"
 
         error_list = []
 
@@ -1127,7 +1127,7 @@ class KhiopsLocalRunner(KhiopsRunner):
                 stacklevel=3,
             )
 
-    def _detect_library_installation_incompatibilities(self, library_root_dir):
+    def _detect_library_installation_incompatibilities(self, library_root_dir_path):
         """Detects known incompatible installations of this library
         in the 3 installation modes see `_infer_khiops_installation_method`
         (binary+pip, conda, conda-based)
@@ -1137,7 +1137,7 @@ class KhiopsLocalRunner(KhiopsRunner):
 
         Parameters
         ----------
-        library_root_dir : PosixPath
+        library_root_dir_path : Path
             path to this current library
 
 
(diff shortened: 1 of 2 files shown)


khiops/core/internals/runner.py
```diff
@@ -1226,11 +1239,19 @@ def _detect_library_installation_incompatibilities(self, library_root_dir):
             # no further check cannot be performed)
             base_dir = _infer_base_dir_for_conda_based_or_pip_installations()
             if len(base_dir) > 0:
+                # Store in separate variable(s) to abstract over
+                # the path casing on Windows
+                base_prefix = sys.base_prefix
```
popescu-v: IMHO we should call `base_prefix` `sys_base_prefix` for coherence with `sys.prefix` -> `sys_prefix`.

khiops/core/internals/runner.py
```diff
@@ -1260,14 +1285,22 @@ def _detect_library_installation_incompatibilities(self, library_root_dir):
                         )
                         error_list.append(error)
                 else:
+                    # Store in separate variable(s) to abstract
+                    # over the path casing on Windows
+                    sys_executable = sys.executable
+                    base_prefix = sys.base_prefix
```
popescu-v: s/base_prefix/sys_base_prefix/ (see comment above in this respect).

khiops/core/internals/runner.py
```diff
@@ -1281,14 +1314,19 @@ def _detect_library_installation_incompatibilities(self, library_root_dir):
...
-                    if not str(library_root_dir).startswith(user_site_packages_dir):
+                    library_root_dir = str(library_root_dir_path)
+                    if platform.system() == "Windows":
+                        # warning : paths are case-insensitive under Windows
+                        library_root_dir = library_root_dir.lower()
+                        user_site_packages_dir = user_site_packages_dir.lower()
+                    if not library_root_dir.startswith(user_site_packages_dir):
                         # the library is not installed on the 'User site'
-                        if not str(library_root_dir).startswith(sys.base_prefix):
+                        if not library_root_dir.startswith(base_prefix):
```
popescu-v: s/base_prefix/sys_base_prefix/

khiops/core/internals/runner.py
```diff
@@ -1240,18 +1261,22 @@ def _detect_library_installation_incompatibilities(self, library_root_dir):
                         # Under Windows, there are two cases :
                         (
                             # for conda-based installations python is inside 'base_dir'
-                            str(Path(sys.executable).parents[0]) != base_dir
+                            # warning : paths are case-insensitive under Windows
+                            str(Path(sys.executable.lower()).parents[0])
```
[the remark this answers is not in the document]
tramora: Indeed I created 2 new variables for clarity sake

khiops/core/internals/runner.py

khiops/core/internals/runner.py
```diff
@@ -1172,10 +1172,18 @@ def _detect_library_installation_incompatibilities(self, library_root_dir):
                 )
                 warning_list.append(warning)
 
+            # Store in separate variable(s) to abstract over the path casing on Windows
+            library_root_dir = str(library_root_dir_path)
+            conda_prefix = os.environ["CONDA_PREFIX"]
```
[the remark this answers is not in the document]
tramora: OK let's got for it. The comparisons will have to be exact then (not using `startswith()` with a prefix any longer). 
One must be careful about the exact folder hierarchy to avoid comparison errors.

PR #558, opened 2026-02-20
Adapt to pandas 3.0

### LINKED ISSUE #536: Adapt to Pandas 3.0.0+
Pandas 3.0.0 has been released on January 21st, 2026. Hence, all environments that use Python >= 3.11 pull Pandas 3.0.0 as a dependency of khiops-python.
However, Pandas 3.0.0 contains breaking changes with respect to previous versions, including Pandas 2.3.3. Most notably, the `StringDType` is used by default, rather than `object`. As the `StringDType` is a Pandas data type, as opposed to `object` which is a NumPy data type, all code assuming a by-default Pandas / NumPy compatibility regarding the Pandas column types breaks now. For example, code like `np.issubdtype(my_pandas_dtype, numpy_dtype)` breaks if `my_pandas_dtype` is a `StringDType`, with `TypeError: Cannot interpret '<StringDtype(storage='python', na_value=nan)>' as a data type`.

tramora:
- For python 3.10, 2.3.3 is still used but with the new pandas StringDtype enabled
- For python 3.11+, the later 3.0.0+ versions are used

Fixes #536 

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,9 @@ pip install -U khiops
 Other installation methods are documented at the [Khiops website][khiops-install].
 
 ### Requirements
-- [Python][python] (>=3.8)
-- [Pandas][pandas] (>=0.25.3)
-- [Scikit-Learn][sklearn] (>=0.22.2)
+- [Python][python] (>=3.10)
+- [Pandas][pandas] (>=2.3.3)
+- [Scikit-Learn][sklearn] (>=1.7.2)
 
 [pandas]: https://pandas.pydata.org
 [sklearn]: https://scikit-learn.org/stable


khiops/sklearn/dataset.py
```diff
@@ -13,7 +13,7 @@
 
 import numpy as np
 import pandas as pd
-import sklearn
+from pandas.core.dtypes.common import is_numeric_dtype, is_string_dtype
```
folmos-at-orange: In the codebase we use this functions without import 
`pd.api.dtypes.common.is_numeric_dtype` and `pd.api.dtypes.common.is_string_dtype`.

Either do that or change everywhere else to use the explicit import.

tests/test_dataset_class.py
```diff
@@ -499,6 +505,12 @@ def test_out_file_from_dataframe_monotable(self):
         out_table = pd.read_csv(out_table_path, sep="\t", dtype={"Title": "string"})
 
         # Cast "Date" columns to datetime as we don't automatically recognize dates
+        # Since pandas 3.0 the default precision for parsing a datetime
+        # is now microseconds (us) instead of nanoseconds (ns)
+        # unless enough precision is given.
+        # Unfortunately only the changelog states this, not the docstring.
+        # To avoid any comparison error in tests
+        # we need to add the required precision to the datetime
         out_table["Date"] = out_table["Date"].astype("datetime64[ns]")
```
popescu-v: Why not just `out_table["Date"] = out_table["Date"].astype("datetime64[us]")` and just stick to the previous value in the fixture?

popescu-v: I think we should have one single consistently-used precision if possible and, to avoid potential loss, I'd stick to nanoseconds, as @folmos-at-orange suggested.

tramora: Your quick test shows the problem @folmos-at-orange, the `assert_frame_equal` fails because of this little difference in precision between the ref date and the out one :  
`datetime64[us]` vs `datetime64[ns]`

khiops/sklearn/dataset.py
```diff
@@ -965,21 +988,23 @@ def __init__(self, name, dataframe, key=None):
...
+        ) and not pd.api.types.is_string_dtype(self.column_ids):
+            previous_type = None
+            for i, column_id in enumerate(self.column_ids):
+                if previous_type is not None and type(column_id) != previous_type:
+                    raise TypeError(
+                        f"Dataframe column ids must be either all integers or "
+                        f"all strings. Column id at index {i} ('{column_id}') is"
+                        f" of type '{type(column_id).__name__}'"
+                    )
+                previous_type = type(column_id)
```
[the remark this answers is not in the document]
tramora: Yes, the loop was removed and the code simplified

tests/test_dataset_class.py
folmos-at-orange: Ok, it is testing code though, so if it works for what we _mean to test_ it is fine.

Issue #244, opened 2024-09-12
Support for Azure Storage


folmos-at-orange:
#### Description
A new Azure Storage file drivers is planned. We should support also Azure Storage for full compatibility with the available official drivers.

#### Questions/Ideas
- New class `AzureStorageResource`
- [Official Python Library](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python?tabs=managed-identity%2Croles-azure-portal%2Csign-in-azure-cli&pivots=blob-storage-quickstart-scratch)

popescu-v:
Closed by PR #571

PR #571, opened 2026-04-10
Support for Azure storage

### LINKED ISSUE #576: Comprehensive integration tests of `FilesystemResource` implementations
#### Description

It would be great to test (via integration tests) comprehensively all the implementations of the `FilesystemResource` abstract class (S3, GCS, Local and Azure).

All the methods must be covered to detect any possible regression.

#### Questions/Ideas
Integration tests already exist for all the implementations. It appears unfortunately that some rarely used methods are not tested.

tramora:
- Among all the Azure storages, Khiops supports only (via its specific driver) "Files" and "Blobs" (Binary Large Objects)
 - The only supported authentication method is currently the `AZURE_STORAGE_CONNECTION_STRING` embedding the account name and account key

Completes #244 and fixes #576

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -138,14 +138,22 @@ jobs:
           echo "Generated AWS configuration..."
           cat  ${GITHUB_WORKSPACE}/.aws/configuration
           /scripts/run_fake_remote_file_servers.sh .  # launch the servers in the background
+
           # Set environment variables for the tests with GCS
           GCS_BUCKET_NAME=data-test-khiops-driver-gcs/khiops_data
           GCS_DRIVER_LOGLEVEL=info  # set to debug for diagnosis
+
           # Set environment variables for the tests with S3
           S3_DRIVER_LOGLEVEL=info  # set to debug for diagnosis
           S3_BUCKET_NAME=s3-bucket
           AWS_SHARED_CREDENTIALS_FILE=${{ github.workspace }}/.aws/credentials
           AWS_CONFIG_FILE=${{ github.workspace }}/.aws/configuration
+
+          # Set environment variables for the tests with Azure
+          AZURE_STORAGE_CONNECTION_STRING='${{ secrets.AZURE_CONNECTION_STRING }}'
+          CLOUD_BLOB_URI_PREFIX=${{ vars.CLOUD_BLOB_URI_PREFIX }}
+          CLOUD_FILE_URI_PREFIX=${{ vars.CLOUD_FILE_URI_PREFIX }}
+                    
           # Persist environment variables for subsequent steps
           echo "GCS_BUCKET_NAME=${GCS_BUCKET_NAME}" >> "$GITHUB_ENV"
           echo "GCS_DRIVER_LOGLEVEL=${GCS_DRIVER_LOGLEVEL}" >> "$GITHUB_ENV"
@@ -153,6 +161,9 @@ jobs:
           echo "S3_BUCKET_NAME=${S3_BUCKET_NAME}" >> "$GITHUB_ENV"
           echo "AWS_SHARED_CREDENTIALS_FILE=${AWS_SHARED_CREDENTIALS_FILE}" >> "$GITHUB_ENV"
           echo "AWS_CONFIG_FILE=${AWS_CONFIG_FILE}" >> "$GITHUB_ENV"
+          echo "AZURE_STORAGE_CONNECTION_STRING=${AZURE_STORAGE_CONNECTION_STRING}" >> "$GITHUB_ENV"
+          echo "CLOUD_BLOB_URI_PREFIX=${CLOUD_BLOB_URI_PREFIX}" >> "$GITHUB_ENV"
+          echo "CLOUD_FILE_URI_PREFIX=${CLOUD_FILE_URI_PREFIX}" >> "$GITHUB_ENV"
       - name: Authenticate to GCP using "Workload Identity Federation"
         if: env.SKIP_EXPENSIVE_TESTS != 'true'
         # For integration tests on GCS we use a real Google account
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -78,39 +78,27 @@ jobs:
         run: |
           CONDA="/root/miniforge3/bin/conda"
 
-          # Native Khiops-based Conda environment, and
-          # `khiops-core`-based Conda environment
-          CONDA_ENVS="py${{ matrix.python-version }} py${{ matrix.python-version }}_conda"
-          for CONDA_ENV in $CONDA_ENVS
-          do
-            mkdir -p -m u+rwx reports/"$CONDA_ENV"
+          # Native Khiops-based Conda environment (to test in a specific python version)
+          CONDA_ENV=py${{ matrix.python-version }}
+          mkdir -p -m u+rwx reports/"$CONDA_ENV"
 
-            # install within the conda environments without activating them
-            $CONDA install -y -n "$CONDA_ENV" unittest-xml-reporting
-            $CONDA install -y -n "$CONDA_ENV" --file test-requirements.txt
-          done
+          # install within the conda environments without activating them
+          $CONDA install -y -n "$CONDA_ENV" unittest-xml-reporting
+          $CONDA install -y -n "$CONDA_ENV" --file test-requirements.txt
       - name: Install khiops-python dependencies
         if: success() || failure()
         run: |
-          # The following git command is required,
-          # as the Git repository is in a directory the current user does not own,
-          # Python versioneer fails to compute the current version correctly otherwise
-          git config --global --add safe.directory $(realpath .)
           CONDA="/root/miniforge3/bin/conda"
-          # Native Khiops-based Conda environment, and
-          # `khiops-core`-based Conda environment
-          CONDA_ENVS="py${{ matrix.python-version }} py${{ matrix.python-version }}_conda"
-          for CONDA_ENV in $CONDA_ENVS
-          do
-            # Since Python 3.13, setuptools is not installed automatically anymore
-            $CONDA install -y -n "$CONDA_ENV" setuptools
+          # Native Khiops-based Conda environment (to test in a specific python version)          
+          CONDA_ENV=py${{ matrix.python-version }}
+          # Since Python 3.13, setuptools is not installed automatically anymore
+          $CONDA install -y -n "$CONDA_ENV" setuptools
 
-            # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
-            $CONDA install -y -n "$CONDA_ENV" tomli
-            $CONDA run --no-capture-output -n "$CONDA_ENV" python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
-            $CONDA install -y -n "$CONDA_ENV" `cat requires.txt`
-            rm -f requires.txt
-          done
+          # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+          $CONDA install -y -n "$CONDA_ENV" tomli
+          $CONDA run --no-capture-output -n "$CONDA_ENV" python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
+          $CONDA install -y -n "$CONDA_ENV" `cat requires.txt`
+          rm -f requires.txt
       - name: Configure Expensive Tests Setting
         # Skip expensive tests by default, unless on the `main-v10` or `main` branches
         if: github.ref != 'main-v10' && github.ref != 'main' && ! inputs.run-expensive-tests
@@ -197,23 +185,17 @@ jobs:
           # version is retrieved
           git config --global --add safe.directory $(realpath .)
           CONDA="/root/miniforge3/bin/conda"
-          # Native Khiops-based Conda environment, and
-          # `khiops-core`-based Conda environment
-          CONDA_ENVS="py${{ matrix.python-version }} py${{ matrix.python-version }}_conda"
-          for CONDA_ENV in $CONDA_ENVS
-          do
-            $CONDA run --no-capture-output -n "$CONDA_ENV" coverage run -m xmlrunner -o "reports/$CONDA_ENV" -v
-            $CONDA run --no-capture-output -n "$CONDA_ENV" coverage report -m
-            $CONDA run --no-capture-output -n "$CONDA_ENV" coverage xml -o "reports/$CONDA_ENV/py-coverage.xml"
-          done
+          # Native Khiops-based Conda environment (to test in a specific python version)          
+          CONDA_ENV=py${{ matrix.python-version }}
+          $CONDA run --no-capture-output -n "$CONDA_ENV" coverage run -m xmlrunner -o "reports/$CONDA_ENV" -v
+          $CONDA run --no-capture-output -n "$CONDA_ENV" coverage report -m
+          $CONDA run --no-capture-output -n "$CONDA_ENV" coverage xml -o "reports/$CONDA_ENV/py-coverage.xml"
       - name: Display Test Reports
         if: success() || failure()
         uses: dorny/test-reporter@v1
         with:
           name: Run Tests ${{ matrix.python-version }}
-          path: >-
-            reports/py${{ matrix.python-version }}/TEST-tests.*.*.xml,
-            reports/py${{ matrix.python-version }}_conda/TEST-tests.*.*.xml
+          path: reports/py${{ matrix.python-version }}/TEST-tests.*.*.xml
           reporter: java-junit
           path-replace-backslashes: 'true'  # Necessary for windows paths
           fail-on-error: 'false'
@@ -225,8 +207,6 @@ jobs:
           path: |-
             reports/py${{ matrix.python-version }}/TEST-tests.*.*.xml
             reports/py${{ matrix.python-version }}/py-coverage.xml
-            reports/py${{ matrix.python-version }}_conda/TEST-tests.*.*.xml
-            reports/py${{ matrix.python-version }}_conda/py-coverage.xml
             tests/resources/scenario_generation/*/ref/*._kh
             tests/resources/scenario_generation/*/output/*._kh
             tests/resources/*/output_reports/*.txt
--- a/khiops/core/internals/filesystems.py
+++ b/khiops/core/internals/filesystems.py
@@ -5,12 +5,12 @@
 # see the "LICENSE.md" file for more details.                                        #
 ######################################################################################
 """Classes to interact with local and remote filesystems"""
-
 import json
 import os
 import shutil
 import warnings
 from abc import ABC, abstractmethod
+from functools import cached_property
 from pathlib import Path
 from urllib.parse import urlparse
 
@@ -37,6 +37,18 @@ try:
     gcs_import_error = None
 except ImportError as import_error:
     gcs_import_error = import_error
+
+# Import azure packages if available
+# Delay an ImportError raising to an instantiation of the resource
+try:
+    from azure.core import credentials, utils
+
+    # pylint: disable=redefined-outer-name
+    from azure.storage import blob, fileshare
+
+    azure_import_error = None
+except ImportError as import_error:
+    azure_import_error = import_error
 
 # pylint: enable=invalid-name
 
@@ -76,6 +88,7 @@ def create_resource(uri_or_path):
         - ``file`` or empty: Local filesystem resource
         - ``s3``: Amazon S3 resource
         - ``gs``: Google Cloud Storage resource
+        - ``https``: Azure Storage resource (files or blobs)
 
     Returns
     -------
@@ -94,6 +107,19 @@ def create_resource(uri_or_path):
                 return AmazonS3Resource(uri_or_path)
             elif uri_info.scheme == "gs":
                 return GoogleCloudStorageResource(uri_or_path)
+            elif uri_info.scheme == "https":
+                # Create the corresponding instance of Azure storage resource
+                # based on a well-known name pattern of the uri netloc.
+                # Among all the Azure storages, Khiops supports only
+                # (via its specific driver):
+                # - Files
+                # - Blobs (Binary Large Objects)
+                if uri_info.netloc.endswith(".file.core.windows.net"):
+                    # netloc of a file share
+                    return AzureStorageFileResource(uri_or_path)
+                else:
+                    # netloc of a blob
+                    return AzureStorageBlobResource(uri_or_path)
             elif scheme == "file":
                 # Reject URI if authority is not empty
                 if uri_info.netloc:
@@ -503,7 +529,7 @@ class LocalFilesystemResource(FilesystemResource):
 class GoogleCloudStorageResource(FilesystemResource):
     """Google Cloud Storage Resource
 
-    By default it reads the configuration from standard location.
+    By default, it reads the configuration from standard location.
     """
 
     def __init__(self, uri):
@@ -733,3 +759,307 @@ class AmazonS3Resource(FilesystemResource):
 
 
 # pylint: enable=no-member
+
+
+class AzureStorageResourceMixin:
+    """Azure compatible Storage Resource Mixin
+
+    See `AzureStorageFileResource` and `AzureStorageBlobResource` for more details.
+
+    """
+
+    def __init__(self, uri):
+        """
+        Azure Storage Resource initializer common to Files and Blobs
+        """
+
+        # Stop initialization if Azure modules are not available
+        if azure_import_error is not None:
+            warnings.warn(
+                "Could not import azure modules. "
+                "Make sure you have installed the azure.core, "
+                "azure.storage.fileshare and azure.storage.blob packages to "
+                "access Azure Storage files."
+            )
+            raise azure_import_error
+
+        super().__init__(uri)
+
+        if len(self.uri_parts) < 2:
+            # parts[0] contains the "share name" (for file shares)
+            # or "container name" (for blobs)
+            # parts[1:] contains the remaining of the path
+            raise ValueError(
+                f"Invalid Azure Resource URI '{uri}'. "
+                "The allowed patterns are: \n"
+                "- File : https://<storage-account-name>.file.core.windows.net/"
+                "<share name>/<0 to n folder name(s)>/<file name>\n"
+                "- Blob : https://<storage-account-name>.blob.core.windows.net/"
+                "<container name>/<0 to n virtual folder name(s)>/<blob name>"
+            )
+
+        # Create the authentication object using the connection string
+        connection_string = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+        mappings = utils.parse_connection_string(connection_string)
+        self.credential = credentials.AzureNamedKeyCredential(
+            mappings.get("accountname"), mappings.get("accountkey")
+        )
+
+    def create_child(self, file_name):
+        return create_resource(child_uri_info(self.uri_info, file_name).geturl())
+
+    def create_parent(self, file_name):  # pylint: disable=unused-argument
+        return create_resource(parent_uri_info(self.uri_info).geturl())
+
+    def _uri_parts(self):
+        return [part for part in self.uri_info.path.split("/") if len(part) > 0]
+
+    @cached_property
+    def uri_parts(self):
+        """Build a list of path parts from a path as a str"""
+        return self._uri_parts()
+
+
+class AzureStorageFileResource(AzureStorageResourceMixin, FilesystemResource):
+    """Azure compatible Storage File Resource
+
+    As a prerequisite, the remote storage account and "file share" on Azure
+    MUST have been created by the user.
+
+    For shared Files, the URI pattern of a resource is the following :
+    https://<storage-account-name>.file.core.windows.net/<share name>/...
+                                                <0 to n folder name(s)>/<file name>
+    The first name after the netloc is the "share name" not a simple "folder name".
+
+    By default, this resource reads the configuration from standard location
+    (environment variables for the moment)
+    """
+
+    def __init__(self, uri):
+        """
+        Azure Storage Resource initializer for Files
+        """
+        super().__init__(uri)
+
+        share_url = (
+            f"{self.uri_info.scheme}://{self.uri_info.netloc}/" f"{self.uri_parts[0]}"
+        )
+        # Instantiate a `ShareClient` and attach it to the current instance.
+        self.azure_share_client = fileshare.ShareClient.from_share_url(
+            share_url=share_url, credential=self.credential
+        )
+
+        self.relative_remaining_path = "/".join(self.uri_parts[1:])
+
+        # Instantiate a `ShareFileClient` and attach it to the current instance.
+        self.file_share_client = self.azure_share_client.get_file_client(
+            self.relative_remaining_path
+        )
+
+        # Instantiate a `ShareDirectoryClient` and attach it to the current instance.
+        self.directory_share_client = self.azure_share_client.get_directory_client(
+            self.relative_remaining_path
+        )
+
+    def write(self, data):
+        # In order to be consistent with the other storage implementations,
+        # if the parent directories are missing they are created one by one.
+        relative_folders_hierarchy = os.path.dirname(self.relative_remaining_path)
+        self._create_the_whole_folders_hierarchy(relative_folders_hierarchy)
+
+        self.file_share_client.upload_file(data)
+
+    def exists(self):
+        """Check if the target resource exists (file or directory)"""
+
+        # Calls to `ShareFileClient` and `ShareDirectoryClient` are required
+        # because :
+        # - the check against `file_share_client` is ``False``
+        #   if the target is a directory.
+        # - the check against `directory_share_client` is ``False``
+        #   if the target is a file.
+        return self.file_share_client.exists() or self.directory_share_client.exists()
+
+    def remove(self):
+        """Remove the target resource either it is a file or a directory.
+        If an error occurs an exception is raised.
+        """
+
+        # Attempt to delete a file first; if it fails, attempt to delete a directory;
+        try:
+            self.file_share_client.delete_file()
+        # pylint: disable=broad-exception-caught
+        except Exception as file_delete_error:
+            try:
+                self.directory_share_client.delete_directory()
+            # pylint: disable=broad-exception-caught
+            except Exception as directory_delete_error:
+                raise RuntimeError(
+                    str(file_delete_error) + " " + str(directory_delete_error)
+                ) from None
+
+    def copy_from_local(self, local_path):
+        # In order to be consistent with the other storage implementations,
+        # if the parent directories are missing they are created one by one.
+        relative_folders_hierarchy = os.path.dirname(self.relative_remaining_path)
+        self._create_the_whole_folders_hierarchy(relative_folders_hierarchy)
+
+        self.file_share_client.upload_file(read(local_path))
+
+    def copy_to_local(self, local_path):
+        data = self.file_share_client.download_file()
+        with open(local_path, "wb") as output_file:
+            data.readinto(output_file)
+
+    def list_dir(self):
+        """List the files (not the directories) of the current directory
+        Notes:
+            This is not a recursive listing operation
+        """
+
+        return [
+            item["name"]
+            for item in self.directory_share_client.list_directories_and_files()
+            if not item["is_directory"]
+        ]
+
+    def make_dir(self):
+        # To avoid any exception while attempting to create an existing directory,
+        # that would occur otherwise, its existence is first checked
+
+        if not self.directory_share_client.exists():
+            self.azure_share_client.create_directory(self.relative_remaining_path)
+
+    def read(self, size=None):
+        return self.file_share_client.download_file(length=size).readall()
+
+    def _create_the_whole_folders_hierarchy(self, relative_folders_hierarchy):
+        """
+        Create the whole folders hierarchy in checking and
+        creating each parent if needed
+        """
+        parts = relative_folders_hierarchy.split("/")
+        current_hierarchy = ""
+        for current_folder in parts:
+            current_hierarchy += current_folder + "/"
+            directory_share_client = self.azure_share_client.get_directory_client(
+                current_hierarchy
+            )
+            if not directory_share_client.exists():
+                self.azure_share_client.create_directory(current_hierarchy)
+
+
+class AzureStorageBlobResource(AzureStorageResourceMixin, FilesystemResource):
+    """Azure compatible Storage Blob Resource
+
+    As a prerequisite, the remote storage account and the blob "container" on Azure
+    MUST have been created by the user.
+
+    For blobs, the URI pattern of a resource is the following:
+    https://<storage-account-name>.blob.core.windows.net/<container name>/...
+                                        <0 to n virtual folder name(s)>/<blob name>
+    The "virtual folder names" are part of the blob name but can help simulate
+    a folder hierarchy.
+    The first name after the netloc is the "container name" not a simple
+    "virtual folder name".
+
+    By default, this resource reads the configuration from standard location
+    (environment variables for the moment)
+    """
+
+    def __init__(self, uri):
+        """
+        Azure Storage Resource initializer for Blobs
+        """
+        super().__init__(uri)
+
+        # For blobs : most of the time, there is no need to distinguish
+        # between `container_url`
+        # and the remaining parts of the path because the `BlobClient` knows
+        # how to handle with absolute paths.
+        # But for blobs listing only, a `ContainerClient` is needed.
+        # Therefore, the `container_url` must still be built.
+        container_url = (
+            f"{self.uri_info.scheme}://"
+            f"{self.uri_info.netloc}/"
+            f"{self.uri_parts[0]}"
+        )
+
+        # Instantiate a `ContainerClient` and attach it to the current instance.
+        self.azure_blob_container_client = blob.ContainerClient.from_container_url(
+            container_url=container_url, credential=self.credential
+        )
+
+        # Instantiate a `BlobClient` and attach it to the current instance.
+        self.azure_blob_client = blob.BlobClient.from_blob_url(
+            blob_url=self.uri_info.geturl(), credential=self.credential
+        )
+
+        # For blobs the client will manage most of the time with absolute paths
+        # except when listing blobs of a "virtual folder" where is field this used.
+        self.relative_remaining_path = "/".join(self.uri_parts[1:])
+
+    def read(self, size=None):
+        return self.azure_blob_client.download_blob(length=size).readall()
+
+    def write(self, data):
+        # In order to be consistent with the other drivers,
+        # this method will overwrite the destination blob if it exists.
+        # This is not the default behavior.
+        self.azure_blob_client.upload_blob(data, overwrite=True)
+
+    def exists(self):
+        # As the virtual folder names are part of the blob name,
+        # there is no such test like checking a virtual folder existence.
+        # Only a blob existence is checked here.
+        return self.azure_blob_client.exists()
+
+    def remove(self):
+        # As the virtual folder names are part of the blob name,
+        # there is no such action like removing a virtual folder.
+        # Only a blob removal is done here.
+        self.azure_blob_client.delete_blob()
+
+    def copy_from_local(self, local_path):
+        # In order to be consistent with the other drivers,
+        # this method will overwrite the destination blob if it exists.
+        # This is not the default behavior.
+        # Moreover, if the virtual folders hierarchy does not exist
+        # it is created automatically as it is part of the blob name.
+        self.azure_blob_client.upload_blob(read(local_path), overwrite=True)
+
+    def copy_to_local(self, local_path):
+        data = self.azure_blob_client.download_blob()
+        with open(local_path, "wb") as output_file:
+            data.readinto(output_file)
+
+    def list_dir(self):
+        """
+        List the files of the current directory
+        """
+        # By default, the sdk will list all the blobs belonging to the container.
+        # An extract filter is then required to simulate a directory listing.
+        # Moreover, the virtual folders hierarchy must be deleted from the result.
+        virtual_folder_hierarchy = self.relative_remaining_path
+        if not virtual_folder_hierarchy.endswith("/"):
+            # Normalize the folder hierarchy with a trailing '/'
+            virtual_folder_hierarchy += "/"
+        len_virtual_folder_hierarchy = len(virtual_folder_hierarchy)
+        return [
+            item[len_virtual_folder_hierarchy:]
+            for item in self.azure_blob_container_client.list_blob_names(
+                # Keep only the blobs belonging to the "virtual folder"
+                name_starts_with=self.relative_remaining_path,
+            )
+        ]
+
+    def make_dir(self):
+        warnings.warn(
+            "'make_dir' is a non-operation on Azure Storage for Blobs. "
+            "See the documentation at "
+            "https://learn.microsoft.com/en-us/rest/api/storageservices/"
+            "operations-on-containers and "
+            "https://learn.microsoft.com/en-us/rest/api/storageservices/"
+            "naming-and-referencing-containers--blobs--and-metadata#blob-names "
+            "(a virtual hierarchy can be created in naming blobs)"
+        )
--- a/packaging/docker/khiopspydev/Dockerfile.debian
+++ b/packaging/docker/khiopspydev/Dockerfile.debian
@@ -49,6 +49,7 @@ RUN true \
 ARG PYTHON_VERSIONS
 ARG KHIOPS_GCS_DRIVER_REVISION
 ARG KHIOPS_S3_DRIVER_REVISION
+ARG KHIOPS_AZURE_DRIVER_REVISION
 
 # Install Conda packages
 # Use `rc` label for alpha or RC khiops-core pre-releases
@@ -73,7 +74,9 @@ RUN true \
         # remote files drivers installed in the conda environment \
         $CONDA install -y -n py${version}_conda \
         khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
-        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION}; \
+        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION} \
+        # XXX hardcoded version because the latest version is not released on conda-forge \
+        khiops-driver-azure==0.0.5; \
     done; \
     # Install Khiops from a different major version in a dedicated conda environment \
     # The python interpreter version of the base environment is used as no specific version is given \
@@ -94,9 +97,11 @@ RUN true \
     # Force the installation of the debian 12 versions (bookworm)
     && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_REVISION}/khiops-driver-gcs_${KHIOPS_GCS_DRIVER_REVISION}-1-bookworm.amd64.deb \
     && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_REVISION}/khiops-driver-s3_${KHIOPS_S3_DRIVER_REVISION}-1-bookworm.amd64.deb \
-    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb || true) \
+    # XXX There is a typo in the latest release tag. Until it is fixed, we need to use the hardcoded folder name
+    && wget -O khiops-azure.deb https://github.com/KhiopsML/khiopsdriver-azure/releases/download/0.0.7/khiops-driver-azure_${KHIOPS_AZURE_DRIVER_REVISION}-1-bookworm.amd64.deb \
+    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb khiops-azure.deb || true) \
     && apt-get -f -y install \
-    && rm -f khiops-gcs.deb khiops-s3.deb \
+    && rm -f khiops-gcs.deb khiops-s3.deb khiops-azure.deb \
     && true
 
 FROM ghcr.io/khiopsml/khiops-server:${SERVER_REVISION} AS server
--- a/packaging/docker/khiopspydev/Dockerfile.debian
+++ b/packaging/docker/khiopspydev/Dockerfile.debian
@@ -44,44 +44,28 @@ RUN true \
   && rm -rf ./khiops \
   && true
 
-# set up all the supported Python environments under conda (for the unit tests)
+# set up all the supported Python environments under Conda (for the unit tests)
 # relying on a variable containing all the versions
 ARG PYTHON_VERSIONS
 ARG KHIOPS_GCS_DRIVER_REVISION
 ARG KHIOPS_S3_DRIVER_REVISION
 ARG KHIOPS_AZURE_DRIVER_REVISION
 
-# Install Conda packages
-# Use `rc` label for alpha or RC khiops-core pre-releases
+# Initialize all the Conda environments
 RUN true \
   && export CONDA="/root/miniforge3/bin/conda" \
-  && /bin/bash -c 'if [[ $(echo ${KHIOPS_REVISION} | grep -E ".*-(a|rc)\.[0-9]+") ]]; then \
-       export RC_LABEL="conda-forge/label/rc::"; \
-     else \
-       export RC_LABEL=""; \
-     fi; \
+  && /bin/bash -c ' \
     # Update conda to the latest version \
     $CONDA update -n base conda; \
     for version in ${PYTHON_VERSIONS}; \
     do \
-        CONDA_ENVS="py${version} py${version}_conda"; \
-        for CONDA_ENV in $CONDA_ENVS; \
-        do \
-            $CONDA create -y -n $CONDA_ENV python=${version}; \
-        done; \
-        # khiops core \
-        $CONDA install -y -n py${version}_conda "${RC_LABEL}"khiops-core==$(echo ${KHIOPS_REVISION} | tr -d "-") ; \
-        # remote files drivers installed in the conda environment \
-        $CONDA install -y -n py${version}_conda \
-        khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
-        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION} \
-        # XXX hardcoded version because the latest version is not released on conda-forge \
-        khiops-driver-azure==0.0.5; \
+        $CONDA create -y -n py${version} python=${version}; \
     done; \
     # Install Khiops from a different major version in a dedicated conda environment \
+    # for a specific test regarding compatibility test \
     # The python interpreter version of the base environment is used as no specific version is given \
     $CONDA create -y -n py3_khiops10_conda; \
-    $CONDA install -y -n py3_khiops10_conda khiops::khiops-core==10.3.1; \
+    $CONDA install -y -n py3_khiops10_conda khiops-core==10.3.2; \
     ' \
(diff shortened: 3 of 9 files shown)


tramora:
The automated tests on conda environments currently fail because of an outdate version of the `azure-storage-file-share` package in the `conda-forge` channel. This PR https://github.com/conda-forge/azure-storage-file-share-feedstock/pull/9 should fix the issue but it is blocked for the moment.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +754,358 @@ def create_parent(self):
 
 
 # pylint: enable=no-member
+
+
+class AzureStorageResourceMixin:
+    """Azure compatible Storage Resource Mixin
+
+    Among all the Azure storages, Khiops supports only (via its specific driver) :
```
folmos-at-orange: Remove the space before `:`.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +754,358 @@ def create_parent(self):
...
+        - Files
+        - Blobs (Binary Large Objects)
+
+    For shared Files, the URI pattern of a resource is the following :
```
folmos-at-orange: Remove the space before `:`.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +754,358 @@ def create_parent(self):
...
+            data = self.azure_blob_client.download_blob()
+            data.readinto(output_file)
+
+    def list_dir(self):
+        """
+        Notes:
+             by default, the sdk will list all the blobs belonging to the container.
+             An extract filter is then required to simulate a directory listing.
+             Moreover, the virtual folders hierarchy must be deleted from the result.
+        """
```
folmos-at-orange, line 1039: I think this should be a code comment not a docstring, because it is an implementation detail.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +754,358 @@ def create_parent(self):
...
+
+    def copy_from_local(self, local_path):
+        """
+        Notes:
+             in order to be consistent with the other drivers,
+             this method will overwrite the destination blob if it exists.
+             This is not the default behavior.
+             Moreover, if the virtual folders hierarchy does not exist
+             it is created automatically as it is part of the blob name.
+        """
```
folmos-at-orange: I think this should be a code comment not a docstring, because it is an implementation detail.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +754,358 @@ def create_parent(self):
...
+        self.azure_blob_client.upload_blob(data, overwrite=True)
+
+    def exists(self):
+        """
+        Notes:
+             as the virtual folder names are part of the blob name,
+             there is no such test like checking a virtual folder existence.
+             Only a blob existence is checked here.
+        """
```
folmos-at-orange: Same as below: This is a comment.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +754,358 @@ def create_parent(self):
...
+        return self.azure_blob_client.exists()
+
+    def remove(self):
+        """
+        Notes:
+             as the virtual folder names are part of the blob name,
+             there is no such action like removing a virtual folder.
+             Only a blob removal is done here.
+        """
```
folmos-at-orange: Same as below: This is a comment.

tests/test_remote_access.py
```diff
@@ -77,13 +108,23 @@ def init_remote_bucket(cls, bucket_name=None, proto=None):
...
+                except:  # pylint: disable=broad-exception-caught
+                    pass
                 fs.copy_to_local(
-                    f"{proto}://{bucket_name}/khiops-cicd/samples/{file}", "/tmp/dummy"
+                    f"{proto}://{bucket_name_or_storage_prefix}/khiops-cicd/"
+                    f"samples/{file}",
+                    "/tmp/dummy",
                 )
+                # We cannot use `unittest.TestCase.assertTrue` in a class method
+                assert os.path.isfile("/tmp/dummy")
```
folmos-at-orange: - I don't get why this new code was added
- If it stays I'd refactor it as:
  - Before the loop create `/tmp/khiops-dummy`
  - Save each file to `/tmp/khiops-dummy/{file}`
  - Do not use `assert` as it is not a programming error, I think it should be something like `if not os.path.isfile(...): raise RuntimeError`
  - Clean `/tmp/khiops-dummy` afterwards
This avoids the `try/except` with `pass` in the middle.

tests/test_remote_access.py
```diff
@@ -312,11 +346,6 @@ def tearDownClass(cls):
         if s3_config_exists():
             kh.get_runner().__init__()
 
-    def should_skip_in_a_conda_env(self):
-        # The S3 driver is now released for conda too.
-        # No need to skip the tests any longer in a conda environment
-        return False
-
     def config_exists(self):
```
folmos-at-orange, line 320: I think these elimination should go in a different commit.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +754,358 @@ def create_parent(self):
 
 
 # pylint: enable=no-member
+
+
+class AzureStorageResourceMixin:
+    """Azure compatible Storage Resource Mixin
```
folmos-at-orange, line 765: I think all these docs should go in the classes `AzureStorage*Resource`.

In the docstring of this class they should point to those, or it could be duplicated.

Also I think we should add an URL to a relevant Azure storage page.

folmos-at-orange (changes requested):

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Corrected

khiops/core/internals/filesystems.py
tramora: Corrected

khiops/core/internals/filesystems.py
tramora: Corrected

khiops/core/internals/filesystems.py
tramora: Corrected

tests/test_remote_access.py
tramora: During the development phase when `fs.copy_to_local` was an empty method, the test would pass even if the file download was not successful. This part had to be improved to assess the file is correctly downloaded (without any error and with the local file creation).

Following your suggestions I improved a little bit the snippet

tests/test_remote_access.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
 
 
 # pylint: enable=no-member
+
+
+class AzureStorageResourceMixin:
+    """Azure compatible Storage Resource Mixin"""
```
folmos-at-orange: I'd add in the body, ```See `AzureStorageFileResource` and `AzureBlobResource` for more details.```

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+
+        file_share_client = self.azure_share_client.get_file_client(
+            self.relative_remaining_path
+        )
+        file_share_client.upload_file(data)
+
+    def exists(self):
+        """
+        Check if the target resource exists (file or directory)
+        """
```
folmos-at-orange: Move this to the first docstring line to keep the codebase style.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+        # Both clients are required because
+        # - the check against `file_share_client` is ``False``
+        #   if the target is a directory.
+        # - the check against `directory_share_client` is ``False``
+        #   if the target is a file.
+        return file_share_client.exists() or directory_share_client.exists()
+
+    def remove(self):
+        """
+        Remove the target resource either it is a file or a directory.
```
folmos-at-orange: Move this to the first docstring line to keep the codebase style.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+            self.azure_blob_client.upload_blob(input_file.read(), overwrite=True)
+
+    def copy_to_local(self, local_path):
+        with open(local_path, "wb") as output_file:
+            data = self.azure_blob_client.download_blob()
+            data.readinto(output_file)
+
+    def list_dir(self):
+        """
+        List the files of the current directory
```
folmos-at-orange, line 1038: Move this to the first docstring line to keep the codebase style.

folmos-at-orange (approved):
I think this is an opportunity to have better docs on file system.

Currently, very little is documented, and we now have 3 different remote systems of which only one (S3) is well documented.

I'd a section for each remote system in the module's docstring. Each section contains how to configure it in each case, and a URI example.

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Added

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+            raise ValueError(
+                f"Invalid Azure Resource URI '{uri}'. "
+                "The allowed patterns are: \n"
+                "- File : https://<storage-account-name>.file.core.windows.net/"
+                "<share name>/<0 to n folder name(s)>/<file name>\n"
+                "- Blob : https://<storage-account-name>.blob.core.windows.net/"
+                "<container name>/<0 to n virtual folder name(s)>/<blob name>"
+            )
+
+        if __class__.is_netloc_of_file_share(self.uri_info.netloc):
```
popescu-v: This if / else should be split and dispatched to the `__init__` methods of the relevant Azure storage class IMHO.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+                "access Azure Storage files."
+            )
+            raise azure_import_error
+
+        super().__init__(uri)
+
+        # Create the authentication object using the connection string
+        connection_string = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+        mappings = utils.parse_connection_string(connection_string)
+        creds = credentials.AzureNamedKeyCredential(
```
popescu-v: I would put this in an attribute named `credentials`, to be reused in the per-storage Azure classes directly (also see the `if / else` split comment below).

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+        self.relative_remaining_path = "/".join(parts[2:])
+
+    def create_child(self, file_name):
+        return create_resource(child_uri_info(self.uri_info, file_name).geturl())
+
+    def create_parent(self, file_name):  # pylint: disable=unused-argument
+        return create_resource(parent_uri_info(self.uri_info).geturl())
+
+    @staticmethod
+    def _splitall(path):
```
popescu-v: I would turn this method in a per-instance method, because it is only used on an instance attribute:
```python
    def _uri_parts(self):
        allparts = []
        path = self.uri_info.path
        while True:
            # Always split on "/" as `path` is a URI
            parts = path.split("/")
            if parts[0] == path:  # sentinel for absolute paths
                allparts.insert(0, parts[0])
                break
            elif parts[1] == path:  # sentinel for relative paths
                allparts.insert(0, parts[1])
                break
            else:
                path = parts[0]
                allparts.insert(0, parts[1])
        return allparts
```

Then, I would also create a `uri_parts` property:
```python
@property
def uri_parts(self):
    return self._uri_parts()
```

and use it instead of `parts`, in the per-storage type Azure classes (also see the `if / else` split comment).

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+            elif parts[1] == path:  # sentinel for relative paths
+                allparts.insert(0, parts[1])
+                break
+            else:
+                path = parts[0]
+                allparts.insert(0, parts[1])
+        return allparts
+
+    @staticmethod
+    def is_netloc_of_file_share(netloc):
```
popescu-v: Once the `if / else` in the mixin's `__init__` has been split to the per-storage classes, this method is no longer needed: in `create_resource` (its only other usage besides the `if / else` mentioned above), we can use `netloc.endswith("...")` directly.

khiops/core/internals/filesystems.py
```diff
@@ -94,6 +107,18 @@ def create_resource(uri_or_path):
...
             elif uri_info.scheme == "gs":
                 return GoogleCloudStorageResource(uri_or_path)
+            elif uri_info.scheme == "https":
+                # Create the corresponding instance of Azure storage resource
+                # based on a well-known name pattern of the uri netloc.
+                # Among all the Azure storages, Khiops supports only
+                # (via its specific driver):
+                # - Files
+                # - Blobs (Binary Large Objects)
+                if AzureStorageResourceMixin.is_netloc_of_file_share(uri_info.netloc):
```
popescu-v: Use `netloc.endswith` directly (see other comments related to this).

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+            self.relative_remaining_path
+        )
+        try:
+            directory_share_client.delete_directory()
+            # If the deletion of the directory succeeds then return immediately
+            return
+        except Exception as delete_error:  # pylint: disable=broad-exception-caught
+            delete_directory_error = delete_error
+
+        raise delete_directory_error or delete_file_error
```
popescu-v: I would rewrite all this method thus:
```python
        file_share_client = self.azure_share_client.get_file_client(
            self.relative_remaining_path
        )

        # Attempt to delete the file first; if this fails, attempt to delete the directory; fail with the Azure
        # exception if this fails
        try:
            file_share_client.delete_file()
        except Exception:  # pylint: disable=broad-exception-caught
            directory_share_client = self.azure_share_client.get_directory_client(
                self.relative_remaining_path
            )
            directory_share_client.delete_directory()
```

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+        # In order to be consistent with the other storage implementations,
+        # if the parent directories are missing they are created one by one.
+        relative_folders_hierarchy = os.path.dirname(self.relative_remaining_path)
+        self._create_the_whole_folders_hierarchy(relative_folders_hierarchy)
+
+        file_share_client = self.azure_share_client.get_file_client(
+            self.relative_remaining_path
+        )
+        with open(local_path) as input_file:
+            file_share_client.upload_file(input_file.read())
```
popescu-v: Why not use `file_share_client.upload_file(read(local_path))` directly?

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+
+    def copy_to_local(self, local_path):
+        file_share_client = self.azure_share_client.get_file_client(
+            self.relative_remaining_path
+        )
+        with open(local_path, "wb") as output_file:
+            data = file_share_client.download_file()
```
popescu-v: Move this line above the `with` (thus out of its scope)?

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+        self.azure_blob_client.delete_blob()
+
+    def copy_from_local(self, local_path):
+        # In order to be consistent with the other drivers,
+        # this method will overwrite the destination blob if it exists.
+        # This is not the default behavior.
+        # Moreover, if the virtual folders hierarchy does not exist
+        # it is created automatically as it is part of the blob name.
+        with open(local_path) as input_file:
+            self.azure_blob_client.upload_blob(input_file.read(), overwrite=True)
```
popescu-v: Why not use `self.azure_blob_client.upload_blob(read(local_path), overwrite=True)` directly?

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +758,333 @@ def create_parent(self):
...
+
+    def copy_to_local(self, local_path):
+        with open(local_path, "wb") as output_file:
+            data = self.azure_blob_client.download_blob()
```
popescu-v: I'd move this line above the `with` line (thus out of  its scope).

packaging/docker/khiopspydev/Dockerfile.debian
```diff
@@ -74,6 +75,8 @@ RUN true \
         $CONDA install -y -n py${version}_conda \
         khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
         khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION}; \
+        # hardcoded version because the latest version is not released on conda-forge \
```
popescu-v: Precede this comment text with `XXX` in order to better locate this for future removal.

packaging/docker/khiopspydev/Dockerfile.debian
```diff
@@ -94,9 +97,11 @@ RUN true \
     # Force the installation of the debian 12 versions (bookworm)
     && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_REVISION}/khiops-driver-gcs_${KHIOPS_GCS_DRIVER_REVISION}-1-bookworm.amd64.deb \
     && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_REVISION}/khiops-driver-s3_${KHIOPS_S3_DRIVER_REVISION}-1-bookworm.amd64.deb \
-    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb || true) \
+    # There is a typo in the latest release tag. Until it is fixed, we need to use the hardcoded folder name
```
popescu-v: Precede comment text with `XXX `.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -72,6 +73,8 @@ RUN true \
         $CONDA install -y -n py${version}_conda \
         khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
         khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION}; \
+        # hardcoded version because the latest version is not released on conda-forge \
```
popescu-v: Precede comment with `XXX`.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -91,9 +94,11 @@ RUN true \
     && if [ -f /etc/os-release ]; then . /etc/os-release; fi \
     && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_REVISION}/khiops-driver-gcs_${KHIOPS_GCS_DRIVER_REVISION}-1-${VERSION_CODENAME}.amd64.deb \
     && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_REVISION}/khiops-driver-s3_${KHIOPS_S3_DRIVER_REVISION}-1-${VERSION_CODENAME}.amd64.deb \
-    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb || true) \
+    # There is a typo in the latest release tag. Until it is fixed, we need to use the hardcoded folder name
```
popescu-v: Precede comment with `XXX`.

tests/test_remote_access.py
```diff
@@ -77,13 +108,22 @@ def init_remote_bucket(cls, bucket_name=None, proto=None):
...
                 )
+
                 # symmetric call to ensure the upload was OK
                 fs.copy_to_local(
-                    f"{proto}://{bucket_name}/khiops-cicd/samples/{file}", "/tmp/dummy"
+                    f"{proto}://{bucket_name_or_storage_prefix}/khiops-cicd/"
+                    f"samples/{file}",
+                    f"/tmp/khiops-{proto}-dummy",
                 )
+                # We cannot use `unittest.TestCase.assertTrue` in a class method
```
popescu-v: I'd remove this comment from here.

popescu-v (changes requested):
See the comments.

khiops/core/internals/filesystems.py
tramora: I changed this

khiops/core/internals/filesystems.py
tramora: I changed this. I saved the credentials into `self.credential` (singular) to avoid name clash with the package name. It seems better than `creds` and the Azure API already accepts a kwarg name `credential`

khiops/core/internals/filesystems.py
tramora: I changed this

khiops/core/internals/filesystems.py
tramora: Removed

khiops/core/internals/filesystems.py
tramora: Modified

packaging/docker/khiopspydev/Dockerfile.debian
tramora: Modified

packaging/docker/khiopspydev/Dockerfile.debian
tramora: Modified

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: Modified

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: Modified

tests/test_remote_access.py
tramora: Modified

khiops/core/internals/filesystems.py
tramora: Modified

khiops/core/internals/filesystems.py
tramora: Modified but it seems an extra indirection as a `LocalFilesystemResource` would be created and we lose a little bit readability

khiops/core/internals/filesystems.py
tramora: Modified

khiops/core/internals/filesystems.py
tramora: Modified but it seems an extra indirection as a LocalFilesystemResource would be created and we lose a little bit readability

khiops/core/internals/filesystems.py
tramora: Modified for simplicity's sake but there will be a misleading error in the following case (as we don't save any longer the real cause in a variable).

The remote resource is a file and its deletion fails (because of rights or another reason).
We try to remove the file as a directory and a misleading error will raise.

popescu-v: In this case IMHO we need to further use the exception; perhaps its error code and  / or message (I'd have a look here https://github.com/Azure/azure-sdk-for-python/blob/292035457f93385cbcefb777b41821a3beed1c39/sdk/storage/azure-storage-file-share/azure/storage/fileshare/_shared/response_handlers.py#L86 for some details that could help).

Alternatively, if the exception type cannot be stated explicitly, the results of the first exception could be concatenated to the second failure by adding a `try / catch` around directory deletion and concatenating the messages; we can also use `raise ... from None` to avoid having the stack traces of all 3 exceptions raised and keep only the latter one (i.e. the context of the last exception that is raised explicitly in our code); and we can raise a `RuntimeError` with the concatenation of Azure-related exceptions' messages:

```python
        file_share_client = self.azure_share_client.get_file_client(
            self.relative_remaining_path
        )

        # Attempt to delete the file first; if this fails, attempt to delete the directory; fail with the Azure
        # exception if this fails
        try:
            file_share_client.delete_file()
        except Exception as file_deletion_error:  # pylint: disable=broad-exception-caught
            directory_share_client = self.azure_share_client.get_directory_client(
                self.relative_remaining_path
            )
            try:
                directory_share_client.delete_directory()
            except Exception as dir_deletion_error:
                raise RuntimeError(str(file_deletion_error) + " " + str(dir_deletion_error)) from None
```

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+        return create_resource(child_uri_info(self.uri_info, file_name).geturl())
+
+    def create_parent(self, file_name):  # pylint: disable=unused-argument
+        return create_resource(parent_uri_info(self.uri_info).geturl())
+
+    def _uri_parts(self):
+        """Build a list of path parts from a path as a str"""
+        allparts = []
+        path = self.uri_info.path
+        while 1:
```
popescu-v: s/while 1/while True/ (more Pythonic)

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+            parts = os.path.split(path)
```
popescu-v: IMHO, we should always split on `/`, as `path` is an URI. Otherwise, one would split on `\` on Windows and I'm not sure that is what we want here. Cf. the comment in this respect, in the previous round of review (https://github.com/KhiopsML/khiops-python/pull/571#discussion_r3100835604).

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+            # parts[1] contains the "share name"
+            # parts[2:] contains the remaining of the path
+            raise ValueError(
+                f"Invalid Azure Resource URI '{uri}'. "
+                "The allowed patterns are: \n"
+                "- File : https://<storage-account-name>.file.core.windows.net/"
+                "<share name>/<0 to n folder name(s)>/<file name>\n"
+                "- Blob : https://<storage-account-name>.blob.core.windows.net/"
+                "<container name>/<0 to n virtual folder name(s)>/<blob name>"
+            )
```
popescu-v: IMHO, this check should have been kept in the mixin (it's the same for both Blob and FileShare storage types).

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+        """Remove the target resource either it is a file or a directory.
+        If an error occurs an exception is raised.
+        """
+        file_share_client = self.azure_share_client.get_file_client(
+            self.relative_remaining_path
+        )
+        # Attempt to delete a file first, if it fails attempt to delete a directory
+        # eventually fails with the Azure exception.
+        # Note : a misleading but accepted error message can be shown
+        # if a failure occurs while deleting an actual remote file.
```
popescu-v: s/while deleting/while attempting to delete/

Also see the other comment (regarding `raise RuntimeError(...) from None`).

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+            # parts[1] contains the "container name"
+            # parts[2:] contains the remaining of the path
+            raise ValueError(
+                f"Invalid Azure Resource URI '{uri}'. "
+                "The allowed patterns are: \n"
+                "- File : https://<storage-account-name>.file.core.windows.net/"
+                "<share name>/<0 to n folder name(s)>/<file name>\n"
+                "- Blob : https://<storage-account-name>.blob.core.windows.net/"
+                "<container name>/<0 to n virtual folder name(s)>/<blob name>"
+            )
```
popescu-v: Move this check to the mixin (it is the same for both child classes).

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+
+        # For blobs : most of the time, there is no need to distinguish
+        # between `container_url`
+        # and the remaining parts of the path because the `BlobClient` knows
+        # how to handle with absolute paths.
+        # But for blobs listing only, a `ContainerClient` is needed.
+        # Therefore, the `container_url` must still be built.
+        # Warning : parts[0] contains '/'
+        container_url = (
+            f"{self.uri_info.scheme}://" f"{self.uri_info.netloc}/{parts[1]}"
```
popescu-v: Use `self.uri_parts` instead of `parts`.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+        )
+
+        # Instantiate a `BlobClient` and attach it to the current instance.
+        self.azure_blob_client = blob.BlobClient.from_blob_url(
+            blob_url=self.uri_info.geturl(), credential=self.credential
+        )
+
+        # For blobs the client will manage most of the time with absolute paths
+        # except when listing blobs of a "virtual folder" where is field is used.
+        self.relative_remaining_path = "/".join(parts[2:])
```
popescu-v: Use `self.uri_parts` instead of `parts`.

packaging/docker/khiopspydev/Dockerfile.debian
```diff
@@ -94,9 +97,11 @@ RUN true \
     # Force the installation of the debian 12 versions (bookworm)
     && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_REVISION}/khiops-driver-gcs_${KHIOPS_GCS_DRIVER_REVISION}-1-bookworm.amd64.deb \
     && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_REVISION}/khiops-driver-s3_${KHIOPS_S3_DRIVER_REVISION}-1-bookworm.amd64.deb \
-    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb || true) \
+    # xxx There is a typo in the latest release tag. Until it is fixed, we need to use the hardcoded folder name
```
popescu-v: Use capital `XXX`: some editors highlight it when in capital.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -72,6 +73,8 @@ RUN true \
         $CONDA install -y -n py${version}_conda \
         khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
         khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION}; \
+        # xxx hardcoded version because the latest version is not released on conda-forge \
```
popescu-v: Use capital `XXX`: some editors highlight it when in capital.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -91,9 +94,11 @@ RUN true \
     && if [ -f /etc/os-release ]; then . /etc/os-release; fi \
     && wget -O khiops-gcs.deb https://github.com/KhiopsML/khiopsdriver-gcs/releases/download/${KHIOPS_GCS_DRIVER_REVISION}/khiops-driver-gcs_${KHIOPS_GCS_DRIVER_REVISION}-1-${VERSION_CODENAME}.amd64.deb \
     && wget -O khiops-s3.deb https://github.com/KhiopsML/khiopsdriver-s3/releases/download/${KHIOPS_S3_DRIVER_REVISION}/khiops-driver-s3_${KHIOPS_S3_DRIVER_REVISION}-1-${VERSION_CODENAME}.amd64.deb \
-    && (dpkg -i --force-all khiops-gcs.deb khiops-s3.deb || true) \
+    # xxx There is a typo in the latest release tag. Until it is fixed, we need to use the hardcoded folder name
```
popescu-v: Use capital `XXX`: some editors highlight it when in capital.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+
+    def __init__(self, uri):
+        """
+        Azure Storage Resource initializer for Files
+        """
+        super().__init__(uri)
+
+        # Allow the extraction of the "share name" (for file shares) from the URL.
+        # This name is always located at the first position in the path.
+        parts = self.uri_parts
```
popescu-v: Remove this line.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+                f"Invalid Azure Resource URI '{uri}'. "
+                "The allowed patterns are: \n"
+                "- File : https://<storage-account-name>.file.core.windows.net/"
+                "<share name>/<0 to n folder name(s)>/<file name>\n"
+                "- Blob : https://<storage-account-name>.blob.core.windows.net/"
+                "<container name>/<0 to n virtual folder name(s)>/<blob name>"
+            )
+
+        # Warning : parts[0] contains '/'
+        share_url = f"{self.uri_info.scheme}://{self.uri_info.netloc}/{parts[1]}"
```
popescu-v: Use `self.uri_parts` directly instead of `parts`.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+        # Instantiate a `ShareClient` and attach it to the current instance.
+        # Later, instances of `ShareFileClient` or `ShareDirectoryClient`
+        # will be created.
+        self.azure_share_client = fileshare.ShareClient.from_share_url(
+            share_url=share_url, credential=self.credential
+        )
+        self.relative_remaining_path = "/".join(parts[2:])
```
popescu-v: Use `self.uri_parts` directly instead of `parts`.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,347 @@ def create_parent(self):
...
+
+    def __init__(self, uri):
+        """
+        Azure Storage Resource initializer for Blobs
+        """
+        super().__init__(uri)
+
+        # Allow the extraction of the "container name" (for blobs) from the URL.
+        # This name is always located at the first position in the path.
+        parts = self.uri_parts
```
popescu-v: Remove this line.

packaging/docker/khiopspydev/Dockerfile.debian
```diff
@@ -74,6 +75,8 @@ RUN true \
         $CONDA install -y -n py${version}_conda \
         khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
         khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION}; \
+        # xxx hardcoded version because the latest version is not released on conda-forge \
```
popescu-v: Use capital `XXX`: some editors highlight it when in capital.

popescu-v (changes requested):
A few loose ends:
- finalize mixin -> child classes code move (see relevant comments)
- finalize exception handling on file share / directory deletion;
- sort out some comment details.

khiops/core/internals/filesystems.py
tramora: `os.path.split(path)` purpose is to split between head and tail (final component of a path), `str.split("/")` has a totally different behavior and using it would change the whole algorithm. 

For the border cases (invalid uri), a check is done in the mixin initializer to ensure a valid azure uri is always given.

popescu-v: Right, we should use `path.rsplit("/", 1)` IMHO with special treatment for the "/" path (not kept via rsplit).

khiops/core/internals/filesystems.py
tramora: Agreed. I modified the method accordingly

khiops/core/internals/filesystems.py
tramora: Modified

khiops/core/internals/filesystems.py
tramora: Moved in the mixin

khiops/core/internals/filesystems.py
tramora: Modified

khiops/core/internals/filesystems.py
tramora: Moved in the mixin

khiops/core/internals/filesystems.py
tramora: Changed but a disavantage is that this requires multiple method calls (we know how the property is implemented)

khiops/core/internals/filesystems.py
tramora: Changed but a disavantage is that this requires multiple method calls (we know how the property is implemented)

packaging/docker/khiopspydev/Dockerfile.debian
tramora: Fixed

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: Fixed

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Removed but using the property implies multiple function calls (we know how the property is implemented)

khiops/core/internals/filesystems.py
tramora: Changed but a disavantage is that this requires multiple method calls (we know how the property is implemented)

khiops/core/internals/filesystems.py
tramora: Changed but a disavantage is that this requires multiple method calls (we know how the property is implemented)

khiops/core/internals/filesystems.py
tramora: Removed but using the property implies multiple function calls (we know how the property is implemented)

packaging/docker/khiopspydev/Dockerfile.debian
tramora: Fixed

tests/test_remote_access.py
```diff
@@ -77,13 +108,21 @@ def init_remote_bucket(cls, bucket_name=None, proto=None):
...
                 fs.copy_to_local(
-                    f"{proto}://{bucket_name}/khiops-cicd/samples/{file}", "/tmp/dummy"
+                    f"{proto}://{bucket_name_or_storage_prefix}/khiops-cicd/"
+                    f"samples/{file}",
+                    f"/tmp/khiops-{proto}-dummy",
                 )
+                if not os.path.isfile(f"/tmp/khiops-{proto}-dummy"):
+                    raise RuntimeError(
+                        f"/tmp/khiops-{proto}-dummy must have been created"
+                    )
```
popescu-v, line 127: IMHO `f"/tmp/khiops-{proto}-dummy"` must be cleaned at the end of this check if the exception is not raised.

tests/test_remote_access.py
popescu-v: Closing, as this has been done.

khiops/core/internals/filesystems.py
popescu-v: IMHO, this is done, hence can we close it, @folmos-at-orange ?

khiops/core/internals/filesystems.py
popescu-v: Right. Thus, in order to improve efficiency, we might have `uri_parts` as a cached property, thus only being computed on first call and then cached for subsequent calls:

```python
from functools import cached_property

...
@cached_property
def uri_parts(self):
    return self._uri_parts()
```

khiops/core/internals/filesystems.py
popescu-v: See above.

khiops/core/internals/filesystems.py
popescu-v: See comment above for avoiding the multiple function calls (use the `cached_property` decorator).

khiops/core/internals/filesystems.py
popescu-v: See above.

khiops/core/internals/filesystems.py
popescu-v: See above.

khiops/core/internals/filesystems.py
popescu-v: See above.

khiops/core/internals/filesystems.py
popescu-v: Closing, as done, @folmos-at-orange ?

khiops/core/internals/filesystems.py
popescu-v: Closing as done, @folmos-at-orange ?

popescu-v (changes requested):
See pending comments.

khiops/core/internals/filesystems.py
tramora: I kept the string.split('/') with an extra test and it seems to simplify the method in a one-liner

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

tests/test_remote_access.py
tramora: Fixed

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,319 @@ def create_parent(self):
...
+        )
+
+    def create_child(self, file_name):
+        return create_resource(child_uri_info(self.uri_info, file_name).geturl())
+
+    def create_parent(self, file_name):  # pylint: disable=unused-argument
+        return create_resource(parent_uri_info(self.uri_info).geturl())
+
+    def _uri_parts(self):
+        """Build a list of path parts from a path as a str"""
```
popescu-v: Move the docstring to the public property.

popescu-v (changes requested):
2 minor details, among which a code-related one.

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,319 @@ def create_parent(self):
...
+        Azure Storage Resource initializer for Files
+        """
+        super().__init__(uri)
+
+        share_url = (
+            f"{self.uri_info.scheme}://{self.uri_info.netloc}/" f"{self.uri_parts[0]}"
+        )
+        # Instantiate a `ShareClient` and attach it to the current instance.
+        # Later, instances of `ShareFileClient` or `ShareDirectoryClient`
+        # will be created.
```
popescu-v: Move the creation of these instances from the various methods to the `__init__` method (as long as the argument is `self.relative_remaining_path` (also see comment below).

Or otherwise change the comment to explain why they cannot be created in the `__init__` method.

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,319 @@ def create_parent(self):
...
+        )
+        file_share_client.upload_file(data)
+
+    def exists(self):
+        """Check if the target resource exists (file or directory)"""
+        file_share_client = self.azure_share_client.get_file_client(
+            self.relative_remaining_path
+        )
+        directory_share_client = self.azure_share_client.get_file_client(
+            self.relative_remaining_path
```
popescu-v: Create and set as attributes in the `__init__` method?

khiops/core/internals/filesystems.py
```diff
@@ -733,3 +759,319 @@ def create_parent(self):
...
+        )
+        file_share_client.upload_file(data)
+
+    def exists(self):
+        """Check if the target resource exists (file or directory)"""
+        file_share_client = self.azure_share_client.get_file_client(
+            self.relative_remaining_path
+        )
+        directory_share_client = self.azure_share_client.get_file_client(
```
popescu-v: s/.get_file_client/.get_directory_client/ to get the (root) directory client, right?

popescu-v (changes requested):
Check the remaining comments.
Check failing remote tests in https://github.com/KhiopsML/khiops-python/actions/runs/24792699160, e.g. https://github.com/KhiopsML/khiops-python/actions/runs/24792699160/job/72554199345#step:12:279 and https://github.com/KhiopsML/khiops-python/actions/runs/24792699160/job/72554199345#step:12:294.

khiops/core/internals/filesystems.py
tramora: Yes, better to create them for once in the `__init__`, I did the change

khiops/core/internals/filesystems.py
tramora: instanciation done in the `__init__`

khiops/core/internals/filesystems.py
tramora: Indeed a typo not detected because of https://github.com/KhiopsML/khiops-python/issues/576
Anyway the instanciation is done now in the `__init__`

tramora:
> Check the remaining comments. Check failing remote tests in https://github.com/KhiopsML/khiops-python/actions/runs/24792699160, e.g. https://github.com/KhiopsML/khiops-python/actions/runs/24792699160/job/72554199345#step:12:279 and https://github.com/KhiopsML/khiops-python/actions/runs/24792699160/job/72554199345#step:12:294.

The error message was not very explicit; khiops failed only with a ```fatal error : Command file : Batch mode failure```.

Naturally the **latest** ubuntu dev image didn't have the azure driver (system-wide or Conda) as **the current PR is not yet merged**.

The latest Conda packages for GCS (0.0.16) and Azure (0.0.6) do not exist (compared to their counterpart on github).
And the previous versions are **not installable under Conda** because a (new) dependency incompatibility (fmt).

Anyway the `khiopspydev-ubuntu22.04:11.0.0.244-support-azure-storage` image contains the system-wide drivers that should be always be the fallback drivers during the executions for both "pyXX" (simulated venv) and "pyXX_conda" (Conda environment). **However despite of this, the execution in a Conda environment fails**. I tried to launch locally the container as described in [khiops-core#1009](https://github.com/KhiopsML/khiops/issues/1009)

packaging/docker/khiopspydev/Dockerfile.debian
```diff
@@ -73,7 +74,9 @@ RUN true \
         # remote files drivers installed in the conda environment \
         $CONDA install -y -n py${version}_conda \
         khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
-        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION}; \
+        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION} \
+        # XXX hardcoded version because the latest version is not released on conda-forge \
+        khiops-driver-azure==0.0.5; \
```
popescu-v: Don't hard-code the driver version here, but only in the dev-docker.yml workflow. Drop the `_conda` envs. Just use the native packages.

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -71,7 +72,9 @@ RUN true \
         # remote files drivers installed in the conda environment \
         $CONDA install -y -n py${version}_conda \
         khiops-driver-s3==${KHIOPS_S3_DRIVER_REVISION} \
-        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION}; \
+        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION} \
+        # XXX hardcoded version because the latest version is not released on conda-forge \
+        khiops-driver-azure==0.0.5; \
```
popescu-v: Don't hard-code the driver version here, but only in the dev-docker.yml workflow. Drop the `_conda` envs. Just use the native packages.

packaging/docker/khiopspydev/Dockerfile.debian
tramora: Fixed. We only use the py3.xx Conda environment to simulate a different python environment (without any real Conda package installation)

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: Fixed. We only use the py3.xx Conda environment to simulate a different python environment (without any real Conda package installation)

.github/workflows/tests.yml
```diff
@@ -78,17 +78,14 @@ jobs:
         run: |
           CONDA="/root/miniforge3/bin/conda"
 
-          # Native Khiops-based Conda environment, and
+          # Native Khiops-based Conda environment (to test in a specific python version)
           # `khiops-core`-based Conda environment
```
popescu-v: This line should be dropped, right?

.github/workflows/tests.yml
```diff
@@ -97,20 +94,16 @@ jobs:
           # Python versioneer fails to compute the current version correctly otherwise
           git config --global --add safe.directory $(realpath .)
```
popescu-v: I would try to opportunistically drop these 2 lines, as versioneer is not used anymore since the drop of `setup.py`, hence, the workflow should work without the Git config change now.

.github/workflows/tests.yml
```diff
@@ -197,23 +190,18 @@ jobs:
...
+          $CONDA run --no-capture-output -n "$CONDA_ENV" coverage xml -o "reports/$CONDA_ENV/py-coverage.xml"
       - name: Display Test Reports
         if: success() || failure()
         uses: dorny/test-reporter@v1
         with:
           name: Run Tests ${{ matrix.python-version }}
           path: >-
-            reports/py${{ matrix.python-version }}/TEST-tests.*.*.xml,
-            reports/py${{ matrix.python-version }}_conda/TEST-tests.*.*.xml
+            reports/py${{ matrix.python-version }}/TEST-tests.*.*.xml
```
popescu-v: I think we could use 
```yaml
path: reports/py${{ matrix.python-version }}/TEST-tests.*.*.xml
```
as there is a single path therein.

packaging/docker/khiopspydev/Dockerfile.rocky
```diff
@@ -61,29 +61,22 @@ RUN true \
...
-            $CONDA create -y -n $CONDA_ENV python=${version}; \
-        done; \
-        $CONDA install -y -n py${version}_conda ${RC_LABEL}khiops-core==$(echo ${KHIOPS_REVISION} | tr -d "-") ; \
+        $CONDA create -y -n py${version} python=${version}; \
     done; \
     # Install Khiops from a different major version in a dedicated conda environment \
+    # for a specific test regarding compatibility test \
     # The python interpreter version of the base environment is used as no specific version is given \
     $CONDA create -y -n py3_khiops10_conda; \
     $CONDA install -y -n py3_khiops10_conda khiops::khiops-core==10.3.1; \
```
popescu-v: `khiops::` is not needed in front of the package name AFAIK (hence, it wasn't present in the previous `*_conda` installs before either).

packaging/docker/khiopspydev/Dockerfile.ubuntu
```diff
@@ -42,41 +42,25 @@ RUN true \
...
-        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION} \
-        # XXX hardcoded version because the latest version is not released on conda-forge \
-        khiops-driver-azure==0.0.5; \
+        $CONDA create -y -n py${version} python=${version}; \
     done; \
     # Install Khiops from a different major version in a dedicated conda environment \
+    # for a specific test regarding compatibility test \
     # The python interpreter version of the base environment is used as no specific version is given \
     $CONDA create -y -n py3_khiops10_conda; \
     $CONDA install -y -n py3_khiops10_conda khiops::khiops-core==10.3.1; \
```
popescu-v: I would drop the leading `khiops::` in the package install instruction (see comment above).

packaging/docker/khiopspydev/Dockerfile.debian
```diff
@@ -44,41 +44,25 @@ RUN true \
...
-        khiops-driver-gcs==${KHIOPS_GCS_DRIVER_REVISION} \
-        # XXX hardcoded version because the latest version is not released on conda-forge \
-        khiops-driver-azure==0.0.5; \
+        $CONDA create -y -n py${version} python=${version}; \
     done; \
     # Install Khiops from a different major version in a dedicated conda environment \
+    # for a specific test regarding compatibility test \
     # The python interpreter version of the base environment is used as no specific version is given \
     $CONDA create -y -n py3_khiops10_conda; \
     $CONDA install -y -n py3_khiops10_conda khiops::khiops-core==10.3.1; \
```
popescu-v: I would drop the leading `khiops::` in the package install instruction above (also see the relevant comment below).

popescu-v:
> > Check the remaining comments. Check failing remote tests in https://github.com/KhiopsML/khiops-python/actions/runs/24792699160, e.g. https://github.com/KhiopsML/khiops-python/actions/runs/24792699160/job/72554199345#step:12:279 and https://github.com/KhiopsML/khiops-python/actions/runs/24792699160/job/72554199345#step:12:294.
> 
> The error message was not very explicit; khiops failed only with a `fatal error : Command file : Batch mode failure`.
> 
> Naturally the **latest** ubuntu dev image didn't have the azure driver (system-wide or Conda) as **the current PR is not yet merged**.
> 
> The latest Conda packages for GCS (0.0.16) and Azure (0.0.6) do not exist (compared to their counterpart on github). And the previous versions are **not installable under Conda** because a (new) dependency incompatibility (fmt).
> 
> Anyway the `khiopspydev-ubuntu22.04:11.0.0.244-support-azure-storage` image contains the system-wide drivers that should be always be the fallback drivers during the executions for both "pyXX" (simulated venv) and "pyXX_conda" (Conda environment). **However despite of this, the execution in a Conda environment fails**. I tried to launch locally the container as described in [khiops-core#1009](https://github.com/KhiopsML/khiops/issues/1009)

Noted. However:
- the native drivers are not the fallback drivers in Conda environments, because the drivers are selected according to the value of the `KHIOPS_DRIVERS_PATH` environment variable as defined in `khiops_env`. Its value is specific to the Conda environment `khiops` is launched in;
- however, the error message in case the driver is missing could be more helpful IMHO, rather than just stating that the file is not found; I'd file an issue in the `khiops` repo to this end.

popescu-v (changes requested):
A few remaining comments.

packaging/docker/khiopspydev/Dockerfile.rocky
tramora: Sorry to disagree, for version 10.3.1 the Conda package is still in a khiops channel
image

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: for version 10.3.1 the Conda package is still in a khiops channel

packaging/docker/khiopspydev/Dockerfile.debian
tramora: for version 10.3.1 the Conda package is still in a khiops channel

.github/workflows/tests.yml
tramora: Indeed, fixed.

.github/workflows/tests.yml
tramora: Cleaned

.github/workflows/tests.yml

tramora:
> > > Check the remaining comments. Check failing remote tests in https://github.com/KhiopsML/khiops-python/actions/runs/24792699160, e.g. https://github.com/KhiopsML/khiops-python/actions/runs/24792699160/job/72554199345#step:12:279 and https://github.com/KhiopsML/khiops-python/actions/runs/24792699160/job/72554199345#step:12:294.
> > 
> > 
> > The error message was not very explicit; khiops failed only with a `fatal error : Command file : Batch mode failure`.
> > Naturally the **latest** ubuntu dev image didn't have the azure driver (system-wide or Conda) as **the current PR is not yet merged**.
> > The latest Conda packages for GCS (0.0.16) and Azure (0.0.6) do not exist (compared to their counterpart on github). And the previous versions are **not installable under Conda** because a (new) dependency incompatibility (fmt).
> > Anyway the `khiopspydev-ubuntu22.04:11.0.0.244-support-azure-storage` image contains the system-wide drivers that should be always be the fallback drivers during the executions for both "pyXX" (simulated venv) and "pyXX_conda" (Conda environment). **However despite of this, the execution in a Conda environment fails**. I tried to launch locally the container as described in [khiops-core#1009](https://github.com/KhiopsML/khiops/issues/1009)
> 
> Noted. However:
> 
>     * the native drivers are not the fallback drivers in Conda environments, because the drivers are selected according to the value of the `KHIOPS_DRIVERS_PATH` environment variable as defined in `khiops_env`. Its value is specific to the Conda environment `khiops` is launched in;
> 
>     * however, the error message in case the driver is missing could be more helpful IMHO, rather than just stating that the file is not found; I'd file an issue in the `khiops` repo to this end.

When the conditions described in the issue [khiops-core#1009](https://github.com/KhiopsML/khiops/issues/1009) are not met (which make the process fail far before), khiops-core is explicit enough in telling it cannot open the remote file. Thus there is no need opening a new issue.

packaging/docker/khiopspydev/Dockerfile.rocky
tramora: Agreed : as 10.3.2 was released in conda-forge let's use it

packaging/docker/khiopspydev/Dockerfile.ubuntu
tramora: Agreed : as 10.3.2 was released in conda-forge let's use it

packaging/docker/khiopspydev/Dockerfile.debian
tramora: Agreed : as 10.3.2 was released in conda-forge let's use it

packaging/docker/khiopspydev/Dockerfile.rocky
popescu-v: Then, I'd use the khiops-core:10.3.2 which is on conda-forge.

packaging/docker/khiopspydev/Dockerfile.ubuntu
popescu-v: Then, I'd use the khiops-core:10.3.2 which is on conda-forge.

packaging/docker/khiopspydev/Dockerfile.debian
popescu-v: Then, I'd use the khiops-core:10.3.2 which is on conda-forge.

popescu-v (approved):
LGTM.
Update Azure driver version and remove `# XXX` comments accordingly in the testing infrastructure when this becomes possible.

Issue #578, opened 2026-04-28
Get rid of Conda to check incompatibility with a different major version of khiops


tramora:
#### Description
Currently, in order to check incompatibility with a different major version of khiops,
the Khiops core Conda package is installed in a dedicated Conda environment in the docker files (for debian, ubuntu and rocky)

Snippet : 
```
    # Install Khiops from a different major version in a dedicated conda environment \
    # for a specific test regarding compatibility test \
    # The python interpreter version of the base environment is used as no specific version is given \
    $CONDA create -y -n py3_khiops10_conda; \
    $CONDA install -y -n py3_khiops10_conda khiops::khiops-core==10.3.1;
```

Later in the github workflow `tests.yml` at the step `Test Khiops Integration` this Conda environment is used to detect the major version mismatch.

#### Questions/Ideas

For this specific purpose an official but older khiops docker image from docker hub could be used (`khiopsml/khiops-ubuntu:10.3.2.1`).

A new job using this image would be created in the github workflow `tests.yml`

popescu-v:
IMHO, this would be obsoleted by #572 . Hence, also the related integration tests and the ad-hoc Khiops 10 Conda environment in the dev Docker files should be removed.

tramora:
Obsoleted by #572

Issue #573, opened 2026-04-20
Enhance Linux tests in using the supported python version matrix outside Conda


tramora:
#### Description

Currently, in the `tests.yml` workflow we run the tests against a matrix of conda envs with the supported python versions (from 3.10 to 3.14).

We should do the same for simple python virtual environments for the tests running on Linux

#### Questions/Ideas

In the `khiopsydev-ubuntu` dev image, we will try to install all the supported and alternative python versions (with uv or ppa etc...).
In the workflow the current step `Run Unit & Integration Tests`  would be duplicated into `Run Unit & Integration Tests in Conda` and `Run Unit & Integration Tests in venv`

popescu-v:
`uv` supports this: https://docs.astral.sh/uv/concepts/python-versions/#requesting-a-version. Hence, I'd go for replacing the Conda envs with `uv` venvs in `tests.yaml`. ~But only after `khiops-core` is released as a Pip package on PyPI.~ `khiops-core` is already on Test PyPI.

popescu-v:
The only Conda tests should stay in the `test-conda-forge-package.yml` workflow IMHO. Hence, any relevant Conda environment should be created in that workflow IMHO.

Indeed, Conda was used in `tests.yml` as a practical way of having the full array of supported Python versions. However, the goal of this issue is to supersede Conda by `uv` for this purpose.

PR #582, opened 2026-05-26
Use khiops-core and khiops-driver-* pip packages containing binaries only

### LINKED ISSUE #572: Depend on the `khiops-core` Pip Package
#### Description
The Khiops core binaries (without the GUI) are currently being packaged to be installable via Pip from PyPI.
There is currently an alpha testing version, `11.0.1-a.3` on Test.PyPI: https://test.pypi.org/project/khiops-core/11.0.1a3.
It doesn't work for Python < 3.9 but this should still be OK for the `khiops-python` testing / CI contexts.

The goal of this issue is to make the Khiops Python library able to use the Pip-packaged Khiops core dependency from PyPI.

#### Questions/Ideas
- There are packaging impacts and code impacts:
  - Packaging impact: add `khiops-core==11.0.1a.3` to the `[project.dependencies]` section of the `pyproject.toml` file; to be changed to `11.0.1` once the official `khiops-core` package is published on PyPI.
  - Code impacts:
    - the runner should stop supporting the "binary+pip" installation mode; only "conda", "conda-based" and "pip" should be kept;
    - the runner failure modes would cease to include native binaries vs Python package checks (including version mismatches); in fact, the checks should be very similar to those performed in the "conda" case (as the `khiops-core` package version should also be pinned, as in the Conda package case).
- We need to pay attention to the case of the system-wide pip installation of the `khiops-core` package:
   - should we forbid such setup?
   - does such a setup overwrite any putative previous native installation of the `khiops-core` files?

tramora:
Fixes #572 and #578 

---


tramora:
False sentiment of success : we simulate multiple Python environments using conda (#573). This is very unfortunate as we are not able yet to detect automatically an issue with the new PyPI packages regarding the remote drivers [khiops-core:1041](https://github.com/KhiopsML/khiops/issues/1041). In my opinion this one is a blocking issue.

popescu-v:
> False sentiment of success : we simulate multiple Python environments using conda (#573). This is very unfortunate as we are not able yet to detect automatically an issue with the new PyPI packages regarding the remote drivers [khiops-core:1041](https://github.com/KhiopsML/khiops/issues/1041). In my opinion this one is a blocking issue.

Hence, to properly test this, issue https://github.com/KhiopsML/khiops-python/issues/573 must be addressed first. Right, @tramora ?

.github/workflows/api-docs.yml
```diff
@@ -60,7 +60,9 @@ jobs:
         run: |
           # Install package itself to install the samples datasets
           python -m pip install --upgrade pip
-          python -m pip install --user .
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          # Use of both 'index-url' and '--extra-index-url' options allows indexes prioritization (PEP 766)
+          python -m pip install --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple --user .
```
popescu-v: I am not sure I see why just `python -m pip install --extra-index-url https://test.pypi.org/simple` does not suffice:
- the latest version of `khiops-core` is taken from `test.pypi.org`;
- all other dependencies are taken from `pypi.org`.

Why is this not enough (no need for drivers to build the API docs)?

.github/workflows/dev-docker.yml
```diff
@@ -67,12 +52,15 @@ jobs:
           set -x
           echo "KHIOPS_REVISION=${{ inputs.khiops-revision || env.DEFAULT_KHIOPS_REVISION }}" >> "$GITHUB_ENV"
           echo "IMAGE_INCREMENT=${{ inputs.image-increment || env.DEFAULT_IMAGE_INCREMENT }}" >> "$GITHUB_ENV"
-          echo "KHIOPSDEV_OS_CODENAME=$(echo '${{ matrix.khiopsdev-os }}' | tr -d '0-9.')" >> "$GITHUB_ENV"
+          KHIOPSDEV_OS_CODENAME=$(echo '${{ matrix.khiopsdev-os }}' | tr -d '0-9.')
+          echo "KHIOPSDEV_OS_CODENAME=${KHIOPSDEV_OS_CODENAME}" >> "$GITHUB_ENV"
+          if [[ "${KHIOPSDEV_OS_CODENAME}" == "ubuntu" || "${KHIOPSDEV_OS_CODENAME}" == "debian" ]]; then
```
popescu-v: I would use a Bash `case` here with an error message in the default case.

.github/workflows/pip.yml
```diff
@@ -102,7 +102,9 @@ jobs:
             source khiops-debian-venv/bin/activate
           fi
           pip install --upgrade pip
-          pip install $(ls khiops*.tar.gz)
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          # Use of both 'index-url' and '--extra-index-url' options allows indexes prioritization (PEP 766)
+          pip install --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple $(ls khiops*.tar.gz)
```
popescu-v: See comment above about the `--extra-index-url`.

.github/workflows/quick-checks.yml
```diff
@@ -20,10 +20,14 @@ jobs:
         run: |
           # Extract and install package dependency requirements from metadata
           pip install pip-tools
-          python -m piptools compile -o requirements.txt
+
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          # Use of both 'index-url' and '--extra-index-url' options allows indexes prioritization (PEP 766)
+          pip-compile --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple -o requirements.txt
```
popescu-v: Idem as above.

.github/workflows/quick-checks.yml
```diff
@@ -20,10 +20,14 @@ jobs:
...
 
           # Install dev dependencies
-          pip install -r requirements.txt
+          # Same indexes prioritization as above
+          pip install --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple -r requirements.txt
```
popescu-v: Idem as above.

.github/workflows/tests.yml
```diff
@@ -267,17 +269,20 @@ jobs:
           python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" -s "\n"  > requires.txt
 
           # Install the Python requirements outside a python venv
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          # Use of both 'index-url' and '--extra-index-url' options allows indexes prioritization (PEP 766)
           Get-Content .\requires.txt `
-            | ForEach-Object {python -m pip install $_.toString()}
+            | ForEach-Object {python -m pip install --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple $_.toString()}
```
popescu-v, line 287: Same as above regarding `--extra-index-url`.

.github/workflows/tests.yml
```diff
@@ -267,17 +269,20 @@ jobs:
...
           # Create and activate a python venv
           python -m venv khiops-windows-venv
           khiops-windows-venv\Scripts\Activate.ps1
 
           # Install the Python requirements inside a venv
           # The venv python executable is used here
+          # Same indexes prioritization as above
           Get-Content .\requires.txt `
-            | ForEach-Object {khiops-windows-venv\Scripts\python -m pip install $_.toString()}
+            | ForEach-Object {khiops-windows-venv\Scripts\python -m pip install --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple $_.toString()}
```
popescu-v: Cf. the `--extra-index-url` comments above.

.github/workflows/api-docs.yml
tramora: It was must first change but ... when the package will be on both PyPI and Test PyPI (under the same version or not), as PyPI is the index in the list, it will be used first. That was the rationale of the reordering

.github/workflows/tests.yml
```diff
@@ -392,7 +397,9 @@ jobs:
           # Install tomli for Python < 3.11
           pip install tomli
           python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
-          pip install `cat requires.txt`
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          # Use of both 'index-url' and '--extra-index-url' options allows indexes prioritization (PEP 766)
+          pip install --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple `cat requires.txt`
```
popescu-v: Same comment as above.

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,27 @@ def _child_uri_info(uri_info, child_name):
     return uri_info._replace(path=_child_path(uri_info.path, child_name))
 
 
+def _check_khiops_driver_library(khiops_drivers_path, remote_storage_type):
```
popescu-v: Have `khiops_drivers_path` as a kwarg set to `None` by default.

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,27 @@ def _child_uri_info(uri_info, child_name):
...
+    """
+    Ensures the dynamic library for the Khiops driver exists
+    """
+    if khiops_drivers_path is None or len(khiops_drivers_path) == 0:
+        # We cannot guarantee where the dynamic library is located.
+        # It seems wiser to avoid raising a false error
+        return
+    extension = ".dll" if platform.system() == "Windows" else ".so"
```
popescu-v: I think the DLL extension on MacOS is `.dylib`. Cf. e.g. https://github.com/KhiopsML/khiopsdriver-gcs/blob/943fef618b689d8e1abb05e6feb70e1f8efd1db7/test/plugin_test.cpp#L32.

khiops/core/internals/filesystems.py
popescu-v: Add type and value checks for `remote_storage_type` (as `assert`s, because this is a "private" function by convention) - also see the comment below in this respect.

tramora:
> > False sentiment of success : we simulate multiple Python environments using conda (#573). This is very unfortunate as we are not able yet to detect automatically an issue with the new PyPI packages regarding the remote drivers [khiops-core:1041](https://github.com/KhiopsML/khiops/issues/1041). In my opinion this one is a blocking issue.
> 
> Hence, to properly test this, issue #573 must be addressed first. Right, @tramora ?

I changed the `$CONDA install -y -n env ...` to `$CONDA run -n env pip install ...` in order to really use the PyPI packages. 
But it is not perfect yet as the run will set the `CONDA_PREFIX` environment variable that is checked by `khiops-python` to determine if the installation is done in a 'conda' method or not. 'conda' method and 'pip' one are close but not totally identical

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,27 @@ def _child_uri_info(uri_info, child_name):
...
 
 
+def _check_khiops_driver_library(khiops_drivers_path, remote_storage_type):
+    """
+    Ensures the dynamic library for the Khiops driver exists
+    """
+    if khiops_drivers_path is None or len(khiops_drivers_path) == 0:
+        # We cannot guarantee where the dynamic library is located.
+        # It seems wiser to avoid raising a false error
+        return
```
popescu-v: I would ban `None` or the empty string (or `bytes`) as a `khiops_drivers_path` value: if this function is called, as it is "private" by convention, we can just `assert` that `khiops_drivers_path` is not `None` and it is of type `str` or `bytes` and non-empty at its very beginning: it is the programmer's responsibility to make sure `khiops_drivers_path` has proper type and value before calling the function IMHO.

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,27 @@ def _child_uri_info(uri_info, child_name):
...
+    extension = ".dll" if platform.system() == "Windows" else ".so"
+    absolute_path = os.path.join(
+        khiops_drivers_path,
```
popescu-v: I would write `os.path.abspath(khiops_drivers_path)` instead, to make sure the path is absolute.

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,27 @@ def _child_uri_info(uri_info, child_name):
...
+        "libkhiopsdriver_file_" f"{remote_storage_type}{extension}",
```
popescu-v: Put all this line inside a single f-string: `f"libkhiopsdriver_file_{remote_storage_type}{extension}"`

khiops/core/internals/filesystems.py
```diff
@@ -538,6 +560,9 @@ class GoogleCloudStorageResource(FilesystemResource):
     """
 
     def __init__(self, uri):
+
+        _check_khiops_driver_library(os.environ.get("KHIOPS_DRIVERS_PATH"), "gcs")
```
popescu-v: Only call this function if `KHIOPS_DRIVERS_PATH` is properly set (see comments above regarding this function). Also, note that `KHIOPS_DRIVERS_PATH` is not supposed to be changed by the end-user. A wrong value of it is an upstream bug.

khiops/core/internals/filesystems.py
```diff
@@ -635,11 +660,14 @@ class AmazonS3Resource(FilesystemResource):
     """
 
     def __init__(self, uri):
+
+        _check_khiops_driver_library(os.environ.get("KHIOPS_DRIVERS_PATH"), "s3")
```
popescu-v: Same comment as above.

khiops/core/internals/filesystems.py
```diff
@@ -778,6 +806,8 @@ def __init__(self, uri):
         Azure Storage Resource initializer common to Files and Blobs
         """
 
+        _check_khiops_driver_library(os.environ.get("KHIOPS_DRIVERS_PATH"), "azure")
```
popescu-v: Same comment as above.

khiops/core/internals/runner.py
```diff
@@ -953,20 +953,33 @@ def __init__(self):
...
-        # Check the `khiops_env` script
-        # On Windows native installations, rely on the `KHIOPS_HOME` environment
-        # variable set by the Khiops Desktop Application installer
+        # Search for the `khiops_env` script location
         installation_method = _infer_khiops_installation_method()
-        if platform.system() == "Windows" and installation_method == "binary+pip":
-            # KHIOPS_HOME variable by default
-            if "KHIOPS_HOME" in os.environ:
+        if platform.system() == "Windows" and installation_method == "pip":
+            # On Windows native installations,
```
popescu-v: On Windows, Pip should also install `khiops-core`. Hence, I am not sure I understand why all this comment is still relevant. Windows native installation should be out-of-scope now, shouldn't it?

khiops/core/internals/runner.py
```diff
@@ -953,20 +953,33 @@ def __init__(self):
...
+            )
+            # Hit : use the tool found in the current environment
+            if os.path.exists(probable_khiops_env):
+                khiops_env_path = probable_khiops_env
+            # Miss : KHIOPS_HOME variable evaluation is the fallback
+            elif "KHIOPS_HOME" in os.environ:
                 khiops_env_path = os.path.join(
                     os.environ["KHIOPS_HOME"], "bin", "khiops_env.cmd"
                 )
-            # Raise error if KHIOPS_HOME is not set
```
popescu-v: Why do we still need this? Either `khiops_env.cmd` can be found as expected according to the current Python environment (virtualenv, user-site or system-wide - discouraged), or it cannot. Why rely on `KHIOPS_HOME`?

khiops/core/internals/runner.py
```diff
@@ -953,20 +953,33 @@ def __init__(self):
...
+            # Raise error if no alternative remains
             else:
                 raise KhiopsEnvironmentError(
-                    "No environment variable named 'KHIOPS_HOME' found. "
+                    "No 'khiops_env.cmd' found in current environment nor "
+                    "in 'KHIOPS_HOME' location. "
```
popescu-v: We should drop `KHIOPS_HOME` fallback IMHO (see comments above).

khiops/core/internals/runner.py
```diff
@@ -1029,6 +1042,11 @@ def _initialize_khiops_environment(self):
                 elif var_name == "KHIOPS_MPI_COMMAND":
                     self._mpi_command_args = shlex.split(var_value)
                     os.environ["KHIOPS_MPI_COMMAND"] = var_value
+                # On Windows "KHIOPS_MPI_DLL_PATH" (containing the Intel MPI Library)
+                # must be appended to "PATH" otherwise Khiops wouldn't find it
+                # and fail immediately
+                elif var_name == "KHIOPS_MPI_DLL_PATH":
+                    os.environ["PATH"] = os.environ.get("PATH") + os.pathsep + var_value
```
popescu-v: I would write either `os.pathsep.join([var_value, os.environ.get("PATH")])`, or `f"{var_value}{os.pathsep}{os.environ.get('PATH')}"` and prefer the former (more readable IMHO).

But anyway, I would prepend `var_value` to the path, to make sure the correct MPI path takes precedence over whatever (potentially system-wide) MPI installation might preexist.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 ### Added
 - (`sklearn`) `keep_selected_variables_only` parameter to the predictors (`KhiopsClassifier` and `KhiopsRegressor`)
 - (General) Support for Azure storage
+### Changed
+- *Internals*:
```
popescu-v: I would place this change under `(General)` and rephrase it to something like:
"- (General) Full-Pip installation support: `khiops` now depends on the `khiops-core` and remote storage driver Pip packages"

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
...
+  - Declaration of khiops-core and khiops drivers as dependencies (mandatory or optional)
```
popescu-v: I would also add an entry to the `### Removed` section, along the lines of:
"- Support for OS-native Khiops Core and Pip Khiops Python library installations"

popescu-v (changes requested):
See the comments.

.github/workflows/api-docs.yml
popescu-v: I see. In this case, in order to make sure all _other_ dependencies than `khiops-*`(core and drivers) are taken from pypi.org, IMHO we need to:
1. First install all dependencies except `khiops-*`-ones from pypi.org
2. Install _only_ the `khiops-*` packages from test.pypi.org only (as the only --index-url, no --extra-index-url).

khiops/core/internals/runner.py
tramora: As I commented in the [initial issue](https://github.com/KhiopsML/khiops-python/issues/572#issuecomment-4564462676), on Linux at least, given the 3 levels "venv pip", "system-wide pip", "system-wide native", if the user removes the `khiops-core` package, the khiops in upper level is used instead (`khiops_env` is searched in the available `PATH`).

On Windows, the system-wide native is the equivalent of the "system-wide native" cited above. In my opinion forbidding its use (in ignoring the `KHIOPS_HOME` set by the native installer) will make a behavior difference between the 2 platforms

khiops/core/internals/runner.py
tramora: related to the comment https://github.com/KhiopsML/khiops-python/pull/582#discussion_r3324995212

khiops/core/internals/runner.py
tramora: related to the comment https://github.com/KhiopsML/khiops-python/pull/582#discussion_r3324995212

.github/workflows/api-docs.yml
tramora: I will split into 2 distinct lines then.

There is a difficulty however in a few cases 
- when a `pip install .` is performed in the `khiops-python` folder. (in `api-docs.yml`)
- when the `khiops-python` "tar.gz" artifact is installed (in `pip.yml`)
- on Windows Powershell where the `perl` and `grep` tools are not available

 It seems impossible to perform a on-the-fly filtering (unless doing complex tricks)

.github/workflows/dev-docker.yml
tramora: I changed the syntax, unless I'm wrong, there is no error message to show in the default case as then a specific Dockerfile is used ("Dockerfile.rocky")

popescu-v: > I changed the syntax, unless I'm wrong, there is no error message to show in the default case as then a specific Dockerfile is used ("Dockerfile.rocky")

I thought about explicitly supporting debian, ubuntu (one case), and rocky (second case), with an error for any other codename.

khiops/core/internals/filesystems.py
tramora: You are right, even if I read in different pages that the ".so" was supported first to avoid differences between the "*BSD/Darwin" world and the Linux... then that ".so" was intented only for "Shared Modules" not "Shared Libraries" that shoud be ".dylib" or ".framework". Nevertheless it is implemented as a ".dylib" in our case.

CHANGELOG.md
tramora: Line added even if I'm not 100% sure because of https://github.com/KhiopsML/khiops-python/pull/582#discussion_r3324995212

khiops/core/internals/runner.py
popescu-v: IMHO, we should not support fallback to native installation: if the user removes the `khiops-core` Pip package, then we should detect this and fail right away (I would say, directly in `khiops/__init__.py`): indeed, now the link between the Python library and the binaries is no longer (only) a matter of local runner, but a matter of packaging. Thus, in `khiops/__init__.py`, we could do something like this:

```python
from importlib.metadata import distribution, PackageNotFoundError

try:
    distribution("khiops-core")
except PackageNotFoundError as exc:
    raise KhiopsEnvironmentError(f"The Khiops binaries are not installed properly: ${exc}") from exc
```
I think this should cover the case evoked here. In any case, IMHO the fallback to OS-native packages should not be authorized, because we would run into version compatibility checks etc, which would defeat the purpose of having a "full-pip" or "full-conda" Python experience with Khiops.

khiops/core/internals/runner.py
popescu-v: See comment https://github.com/KhiopsML/khiops-python/pull/582#discussion_r3325814859.

khiops/core/internals/runner.py
popescu-v: See comment https://github.com/KhiopsML/khiops-python/pull/582#discussion_r3325814859.

CHANGELOG.md
popescu-v: See comment https://github.com/KhiopsML/khiops-python/pull/582#discussion_r3325814859.

CHANGELOG.md
tramora: Modified

khiops/core/internals/runner.py
tramora: Improved

khiops/core/internals/filesystems.py
tramora: Improved

khiops/core/internals/filesystems.py
tramora: Improved

khiops/core/internals/filesystems.py
tramora: Improved

khiops/core/internals/filesystems.py
tramora: Improved

khiops/core/internals/filesystems.py
tramora: Improved

khiops/core/internals/filesystems.py
tramora: Improved

khiops/core/internals/filesystems.py
tramora: Improved

tramora: Improved

.github/workflows/tests.yml
tramora: Refactored in splitting in two lines except for Windows (PowerShell) where I do not know yet the tools that would help me

.github/workflows/tests.yml
tramora: for Windows (PowerShell), I do not know yet the tools that would help me split into 2 lines

.github/workflows/tests.yml
tramora: for Windows (PowerShell), I do not know yet the tools that would help me split into 2 lines

.github/workflows/quick-checks.yml
tramora: Refactored in splitting in two lines except for Windows (PowerShell) where I do not know yet the tools that would help me

.github/workflows/quick-checks.yml
tramora: Refactored in splitting in two lines except for Windows (PowerShell) where I do not know yet the tools that would help me

.github/workflows/pip.yml
tramora: Refactored in splitting in two lines except for Windows (PowerShell) where I do not know yet the tools that would help me

.github/workflows/api-docs.yml
tramora: Refactored in splitting in two lines except for Windows (PowerShell) where I do not know yet the tools that would help me

.github/workflows/dev-docker.yml
tramora: Changed but in my opinion it adds a burden to the maintainer (compared to the situation before this PR).
The matrix of all the supported OSes is defined in the same file and the Dockerfile extension was evaluated using each value of the matrix (after suppressing the digits or .). It seems unnecessary to verify that every value is in the same matrix

khiops/core/internals/runner.py
tramora: Agreed. 

I then copied your snippet into `khiops/__init.py` but I had to comment it out immediately until we find a compatible fix.
Indeed in 'conda' and 'conda-based' installation methods where a Conda package is installed (not a PyPI package)
`distribution("khiops-core")` always raises a `PackageNotFoundError`

On Windows 'pip', I also modified the error message if `khiops_env` cannot be found : the user must check the installation of the 'Khiops Python library' not 'Khiops' any longer

khiops/core/internals/runner.py
tramora: Changed

khiops/core/internals/runner.py
tramora: Changed

.github/workflows/api-docs.yml
popescu-v: Another way would be to update the `scripts/extract_dependencies_from_pyptoject_toml.py` script to filter in / out `khiops*` dependencies.

.github/workflows/api-docs.yml
```diff
@@ -60,6 +60,15 @@ jobs:
         run: |
           # Install package itself to install the samples datasets
           python -m pip install --upgrade pip
+          # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+          python -m pip install tomli
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
```
popescu-v: I would consider updating the `extract_dependencies_from_pyproject_toml.py` script to extract either `khiops*`dependencies or the others.

.github/workflows/dev-docker.yml
popescu-v: I find it more legible this way, because more explicit.

.github/workflows/pip.yml
```diff
@@ -102,10 +102,23 @@ jobs:
             source khiops-debian-venv/bin/activate
           fi
           pip install --upgrade pip
+          # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+          pip install tomli
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
+          # First, install all dependencies except khiops-core and khiops-drivers-*
```
popescu-v, line 114: I would update the `extract_dependencies_from_pyproject_toml.py` script to support partitioning on `khiops*` dependencies (see previous comments in this respect). This would allow for a more homogeneous treatment between Bash and Powershell.

.github/workflows/quick-checks.yml
```diff
@@ -19,11 +19,14 @@ jobs:
...
         run: |
           # Extract and install package dependency requirements from metadata
-          pip install pip-tools
-          python -m piptools compile -o requirements.txt
-
-          # Install dev dependencies
-          pip install -r requirements.txt
+          # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+          pip install tomli
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
```
popescu-v: I would update the Python script to support the khiops* / non-khiops* distinction (see above).

.github/workflows/tests.yml
```diff
@@ -97,7 +97,12 @@ jobs:
           # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
           $CONDA install -y -n "$CONDA_ENV" tomli
           $CONDA run --no-capture-output -n "$CONDA_ENV" python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
-          $CONDA install -y -n "$CONDA_ENV" `cat requires.txt`
+          # Conda is currently used to simulate multiple Python environment, let's avoid installing Conda Packages at all but Pip ones instead
```
popescu-v: I would rephrase this comment to: "`# Conda environments are only used to support multiple Python versions, hence install Pip packages inside`"

.github/workflows/tests.yml
```diff
@@ -97,7 +97,12 @@ jobs:
...
+          $CONDA install -y -n "$CONDA_ENV" pip  # Required otherwise the system pip would be run which will install outside the Conda env
```
popescu-v: I'd put comment before the line and rephrase it to something like "`# Install Pip inside the Conda env to use it instead of the system-wide Pip`".

.github/workflows/tests.yml
```diff
@@ -97,7 +97,12 @@ jobs:
...
+          # First, install all dependencies except khiops-core and khiops-drivers-*
+          $CONDA run -n "$CONDA_ENV" pip install `perl -pe "s/khiops-\S+//g" requires.txt`
```
popescu-v: I would update the Python script to support this (see comments above).

.github/workflows/tests.yml
popescu-v: I would just update the Python script to support extracting either the `khiops*` dependencies or their complement, and use it to generate two different requirements files. This would then allow homogeneous usage across all platforms.

.github/workflows/tests.yml
popescu-v: Update the Python script instead (see above).

.github/workflows/tests.yml
```diff
@@ -392,7 +400,10 @@ jobs:
           # Install tomli for Python < 3.11
           pip install tomli
           python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
-          pip install `cat requires.txt`
+          # First, install all dependencies except khiops-core and khiops-drivers-*
```
popescu-v, line 414: Update the Python script to support this (see above).

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,45 @@ def _child_uri_info(uri_info, child_name):
     return uri_info._replace(path=_child_path(uri_info.path, child_name))
 
 
+def _check_khiops_driver_library(remote_storage_type, khiops_drivers_path=None):
```
popescu-v: On second thought, I think `khiops_drivers_path` should not be `None` and a specific `assert` that it is not should be added. Thus, it would be the programmer's responsibility to ensure that `khiops_drivers_path` is always defined.

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,45 @@ def _child_uri_info(uri_info, child_name):
...
+
+    khiops_drivers_path : `str`
+        Absolute path to the folder containing all the drivers libraries.
+        If a relative path (without the leading '/') is passed,
+        an absolute path is inferred using the current folder which will probably
+        cause an error.
+
+    this method raises a `RuntimeError` if the dynamic library is not found
+    """
+    assert type(remote_storage_type) == str
```
popescu-v: I would also `assert` that `remote_storage_type` and `khiops_drivers_path` are always defined to non-empty strings.

khiops/core/internals/runner.py
popescu-v: Then perhaps move this check to the Pip installation context only (that is, if the installation method is "pip")?

I do not understand you last comment on the Windows 'pip' case: why is it different from the UNIX cases?

khiops/core/internals/runner.py
```diff
@@ -987,7 +986,8 @@ def _initialize_khiops_environment(self):
                 raise KhiopsEnvironmentError(
                     "The 'khiops_env' script not found for the current "
                     f"'{installation_method}' installation method. Make sure "
-                    "you have installed khiops >= 10.2.3. "
+                    "you have installed properly the Khiops Python library "
+                    "and Khiops (if this later was not installed automatically). "
```
popescu-v: s/this later/the latter/

khiops/core/internals/runner.py
```diff
@@ -1088,6 +1095,8 @@ def _initialize_khiops_version(self):
         installation_method = _infer_khiops_installation_method()
 
         # Fail immediately if the major versions differ
+        # It can still occur if the installation method is 'conda' or 'conda-based',
```
popescu-v: Why? In the Conda case, the dependency is also specified, in `meta.yml`, right?

khiops/__init__.py
```diff
@@ -22,10 +22,27 @@
...
 - sklearn: Scikit-Learn estimator classes to learn and use Khiops models
 """
+from importlib.metadata import PackageNotFoundError, distribution
+
+from khiops.core import KhiopsEnvironmentError
 from khiops.core.internals.version import KhiopsVersion
 
 __version__ = "11.0.0.3"
 
+# The current Khiops Python library (when packaged as a PyPI package)
```
popescu-v: I would move this snippet to the runner, in the context of the "pip" installation method.

.github/workflows/pip.yml
popescu-v: Update the Python script directly (see the other comments on this topic).

.github/workflows/quick-checks.yml
popescu-v: Update the Python script directly (see the other comments on this topic).

.github/workflows/quick-checks.yml
popescu-v: Update the Python script directly (see the other comments on this topic).

.github/workflows/tests.yml
popescu-v: Update the Python script directly (see the other comments on this topic).

popescu-v (changes requested):
See pending comments.

khiops/__init__.py
tramora: Moved into the `KhiopsLocalRunner::_detect_library_installation_incompatibilities`...

If the `khiops-core` PyPI package was removed, an error is raised far before in  `KhiopsLocalRunner::_initialize_khiops_environment` except in a very edge case I faced during my tests : `khiops-core` PyPI package was installed system-wide. In this case `KhiopsLocalRunner::_initialize_khiops_environment` runs successfully but the current detection raises an error

.github/workflows/api-docs.yml
tramora: Indeed it's cleaner and portable across the OS. I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/pip.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/quick-checks.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/quick-checks.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/tests.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/tests.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/tests.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

khiops/core/internals/runner.py
tramora: Moved to the `runner.py`

.github/workflows/api-docs.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/pip.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/quick-checks.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/tests.yml

.github/workflows/tests.yml

.github/workflows/tests.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

.github/workflows/tests.yml
tramora: Improved :  I added the options `--exclude-khiops-family` and `--khiops-family-only` to the script

khiops/core/internals/filesystems.py
tramora: Changed

khiops/core/internals/filesystems.py
tramora: Changed

khiops/core/internals/runner.py

tramora:
An issue remains while executing the "expensive tests" and thus running the integration tests for remote files access.
It is tracked on the drivers repository : https://github.com/KhiopsML/khiopsdriver-shared/issues/2

.github/workflows/api-docs.yml
```diff
@@ -60,6 +60,16 @@ jobs:
         run: |
           # Install package itself to install the samples datasets
           python -m pip install --upgrade pip
+          # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+          python -m pip install tomli
+          # First, install all dependencies except khiops-core and khiops-drivers-*
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --exclude-khiops-family > requires-no-khiops.txt          
+          python -m pip install --user `cat requires-no-khiops.txt`
```
popescu-v, line 67: Why not `-r requires-no-khiops.txt` rather than `cat requires-no-khiops.txt` ?

.github/workflows/api-docs.yml
```diff
@@ -60,6 +60,16 @@ jobs:
...
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --khiops-family-only > requires-khiops.txt
+          python -m pip install --user --index-url https://test.pypi.org/simple `cat requires-khiops.txt`
```
popescu-v, line 70: Why not `-r` rather than `cat`?

.github/workflows/pip.yml
```diff
@@ -102,6 +109,16 @@ jobs:
...
           fi
           pip install --upgrade pip
+          # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+          pip install tomli
+          # First, install all dependencies except khiops-core and khiops-drivers-*
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --exclude-khiops-family > requires-no-khiops.txt
+          pip install `cat requires-no-khiops.txt`
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --khiops-family-only > requires-khiops.txt
+          pip install --index-url https://test.pypi.org/simple `cat requires-khiops.txt`
```
popescu-v, line 119: Why not `-r` rather than `cat`?

.github/workflows/pip.yml
```diff
@@ -102,6 +109,16 @@ jobs:
             source khiops-debian-venv/bin/activate
           fi
           pip install --upgrade pip
+          # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+          pip install tomli
+          # First, install all dependencies except khiops-core and khiops-drivers-*
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --exclude-khiops-family > requires-no-khiops.txt
+          pip install `cat requires-no-khiops.txt`
```
popescu-v, line 116: Why not `-r` rather than `cat`?

.github/workflows/quick-checks.yml
```diff
@@ -19,11 +19,15 @@ jobs:
...
-          pip install pip-tools
-          python -m piptools compile -o requirements.txt
-
-          # Install dev dependencies
-          pip install -r requirements.txt
+          # Add homogeneous TOML support (Python >= 3.12 has standard tomllib)
+          pip install tomli
+          # First, install all dependencies except khiops-core and khiops-drivers-*
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --exclude-khiops-family > requires-no-khiops.txt
+          pip install `cat requires-no-khiops.txt`
```
popescu-v, line 26: Why not `-r` rather than `cat`?

.github/workflows/quick-checks.yml
```diff
@@ -19,11 +19,15 @@ jobs:
...
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --khiops-family-only > requires-khiops.txt
+          pip install --index-url https://test.pypi.org/simple `cat requires-khiops.txt`
```
popescu-v, line 29: Why not `-r` rather than `cat`?

.github/workflows/tests.yml
```diff
@@ -96,9 +96,16 @@ jobs:
...
           $CONDA install -y -n "$CONDA_ENV" tomli
-          $CONDA run --no-capture-output -n "$CONDA_ENV" python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
-          $CONDA install -y -n "$CONDA_ENV" `cat requires.txt`
-          rm -f requires.txt
+          # Conda environments are only used to support multiple Python versions, hence install Pip packages inside
+          # Install Pip inside the Conda env to use it instead of the system-wide Pip
+          $CONDA install -y -n "$CONDA_ENV" pip
+          # First, install all dependencies except khiops-core and khiops-drivers-*
+          $CONDA run --no-capture-output -n "$CONDA_ENV" python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --exclude-khiops-family > requires-no-khiops.txt
+          $CONDA run -n "$CONDA_ENV" pip install `cat requires-no-khiops.txt`
```
popescu-v, line 104: Why not `-r` rather than `cat`?

.github/workflows/tests.yml
```diff
@@ -96,9 +96,16 @@ jobs:
...
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          $CONDA run --no-capture-output -n "$CONDA_ENV" python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --khiops-family-only > requires-khiops.txt
+          $CONDA run -n "$CONDA_ENV" pip install --index-url https://test.pypi.org/simple `cat requires-khiops.txt`
```
popescu-v, line 107: Why not `-r` rather than `cat`?

.github/workflows/tests.yml
```diff
@@ -391,9 +408,13 @@ jobs:
 
           # Install tomli for Python < 3.11
           pip install tomli
-          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" > requires.txt
-          pip install `cat requires.txt`
-          rm -f requires.txt
+          # First, install all dependencies except khiops-core and khiops-drivers-*
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --exclude-khiops-family > requires-no-khiops.txt
+          pip install `cat requires-no-khiops.txt`
```
popescu-v, line 416: Why not `-r` rather than `cat`?

.github/workflows/tests.yml
```diff
@@ -391,9 +408,13 @@ jobs:
...
+          # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
+          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --khiops-family-only > requires-khiops.txt
+          pip install --index-url https://test.pypi.org/simple `cat requires-khiops.txt`
```
popescu-v, line 419: Why not `-r` rather than `cat`?

khiops/core/internals/runner.py
```diff
@@ -953,24 +953,24 @@ def __init__(self):
         self._initialize_khiops_environment()
 
     def _initialize_khiops_environment(self):
-        # Check the `khiops_env` script
-        # On Windows native installations, rely on the `KHIOPS_HOME` environment
-        # variable set by the Khiops Desktop Application installer
+        # Search for the `khiops_env` script location
         installation_method = _infer_khiops_installation_method()
-        if platform.system() == "Windows" and installation_method == "binary+pip":
```
popescu-v, line 960: Why not put the `khiops-core` package check in case of Pip installation here?

khiops/core/internals/runner.py
```diff
@@ -1220,13 +1230,31 @@ def _detect_library_installation_incompatibilities(self, library_root_dir_path):
...
+            # are installed system-wide despite all the recommendations)
+            try:
+                distribution("khiops-core")
+            except PackageNotFoundError as exc:
+                error = (
+                    f"The Khiops binaries are not installed properly: {exc}. "
+                    "Re-install the Khiops Python library to automatically install "
+                    "Khiops. Go to https://khiops.org for more information.\n"
+                )
+                error_list.append(error)
```
popescu-v: I would put this check and fail (raise exception) far earlier, when the attempt is made to initialize the Khiops environment, just after the installation method has been detected (see the other comment in this respect, above).

khiops/core/internals/runner.py
popescu-v: s/this latter/the latter/

khiops/__init__.py
popescu-v: I think the environment should not initialize successfully in this case. See comments https://github.com/KhiopsML/khiops-python/pull/582/changes#r3373317679 and https://github.com/KhiopsML/khiops-python/pull/582/changes#r3373326847.

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,45 @@ def _child_uri_info(uri_info, child_name):
...
+    remote_storage_type : `str`
+        Type of the supported storage type
+
+    khiops_drivers_path : `str`
+        Absolute path to the folder containing all the drivers libraries.
+        If a relative path (without the leading '/') is passed,
+        an absolute path is inferred using the current folder which will probably
+        cause an error eventually.
+
+    this method raises a `RuntimeError` if the dynamic library is not found
```
popescu-v: Reformat this line to:
```
Raises
-------
`RuntimeError`
    The dynamic library of the driver is not found.
```

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,45 @@ def _child_uri_info(uri_info, child_name):
...
+    """
+    assert type(remote_storage_type) == str and len(remote_storage_type) > 0
```
popescu-v: The second `assert` implies the first one AFAIU. Hence, I would remove the first assert.

.github/workflows/api-docs.yml
tramora: It is because the script outputs by default the dependencies on a single line but not under the requirements.txt format (multi-lines dependencies). The `cat` method was then used long before this PR. It can be changed if needed.

.github/workflows/api-docs.yml
tramora: cf comment above

.github/workflows/pip.yml
tramora: cf comment above

.github/workflows/pip.yml
tramora: cf comment above

.github/workflows/quick-checks.yml
tramora: cf comment above

.github/workflows/quick-checks.yml
tramora: cf comment above

.github/workflows/tests.yml
tramora: cf comment above

.github/workflows/tests.yml
tramora: cf comment above

.github/workflows/tests.yml
tramora: cf comment above

.github/workflows/tests.yml
tramora: cf comment above

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/__init__.py
tramora: Fixed

khiops/core/internals/runner.py
tramora: Fixed

khiops/core/internals/runner.py

khiops/core/internals/runner.py
tramora: You are right. I removed the whole block of code

khiops/core/internals/filesystems.py
popescu-v: s/the dynamic library/The dynamic library/ (capital "T").

.github/instructions/docker-changes.instructions.md
```diff
@@ -17,23 +17,22 @@ by the `dev-docker.yml` workflow.
...
 
-`Dockerfile.debian` and `Dockerfile.ubuntu` are nearly identical. They diverge
-because Debian 13 remote file driver packages are not available, so Debian
-forces the Debian 12 (bookworm) builds for the GCS, S3, and Azure driver `.deb`
-packages. Any shared change should be applied to both files. There is an open
-TODO to unify them (see the comment at the top of `Dockerfile.debian`).
+`Dockerfile.ubuntu-debian` is a unique Dockerfile to build either a Ubuntu or a Debian target image 
+because there is no difference except for the base image used.
+
+The only reason why the system-wide native Khiops binary (from a `.deb` package) is still installed in the target image
```
popescu-v: IMHO, we can leave that restriction aside if we just install via Pip in a virtualenv, then set the `KHIOPS_CMD` and `KHIOPS_COCLUSTERING_CMD` environment variables accordingly. See: https://github.com/KhiopsML/khiops-server/blob/7050d79fb608c59eed4831838d3753867a2a80f6/server/server.go#L63 and https://github.com/KhiopsML/khiops-server/blob/7050d79fb608c59eed4831838d3753867a2a80f6/server/server.go#L70. I would set these in the CI where applicable.

.github/workflows/dev-docker.yml
```diff
@@ -16,7 +13,7 @@ on:
       khiops-revision:
         type: string
         default: 11.0.0
-        description: Khiops Revision
+        description: Khiops Revision (for tests against KhiopsDockerRunner)
```
popescu-v: I'd use virtualenv and set `KHIOPS_CMD` and `KHIOPS_COCLUSTERING_CMD` (see the previous comment).

khiops/core/internals/filesystems.py
```diff
@@ -127,6 +128,48 @@ def _child_uri_info(uri_info, child_name):
...
+    khiops_drivers_path : `str`
+        Absolute path to the folder containing all the drivers libraries.
+        If a relative path (without the leading '/') is passed,
+        an absolute path is inferred using the current folder which will probably
+        cause an error eventually.
+
+    Raises
+    ------
+    `RuntimeError`
+        the dynamic library of the driver is not found
```
popescu-v: Use capital "T" in "the dynamic".

khiops/core/internals/filesystems.py
```diff
@@ -778,6 +835,12 @@ def __init__(self, uri):
         Azure Storage Resource initializer common to Files and Blobs
         """
 
+        khiops_drivers_path = os.environ.get("KHIOPS_DRIVERS_PATH")
+        if khiops_drivers_path:
+            _check_khiops_driver_library(
+                remote_storage_type="azure", khiops_drivers_path=khiops_drivers_path
+            )
+
```
popescu-v: I'm not sure we should access `KHIOPS_DRIVERS_PATH` directly. Normally, this environment variable name should be abstracted away by the `KhiopsLocalRunner`. I would create specific attribute for accessing it instead, e.g. `khiops_drivers_path`, akin to `khiops_path` and `khiops_coclustering_path`.

khiops/core/internals/runner.py
```diff
@@ -243,11 +243,11 @@ def _infer_khiops_installation_method(trace=False):
        it was not activated previously nor during the execution
        and thus the CONDA_PREFIX environment variable is undefined
        and the path to the `bin` directory inside the conda environment is not in PATH
-    - 'binary+pip' installs the binaries and the shared libraries system-wide
-      but will keep the python libraries
-      in the python system folder
-      or in the Python folder inside the home directory of the user,
-      or in a virtual environment (if one is used)
+    - 'pip' environment containing binaries, shared libraries and the python libraries
```
popescu-v: s/python/Python/

khiops/core/internals/runner.py
```diff
@@ -243,11 +243,11 @@ def _infer_khiops_installation_method(trace=False):
...
+      can either be :
+      - system-wide (strongly discouraged)
+      - or in the Python folder inside the home directory of the user (aka "User site"),
```
popescu-v: s/aka/a.k.a./

khiops/core/internals/runner.py
```diff
@@ -953,24 +953,37 @@ def __init__(self):
...
+            if os.path.exists(probable_khiops_env):
+                khiops_env_path = probable_khiops_env
+            # Raise error otherwise
             else:
                 raise KhiopsEnvironmentError(
-                    "No environment variable named 'KHIOPS_HOME' found. "
-                    "Make sure you have installed Khiops >= 10.2.3. "
+                    "No 'khiops_env.cmd' found in the current environment. "
+                    "Make sure you have installed properly the Khiops Python library "
+                    "and Khiops (if the latter was not installed automatically). "
```
popescu-v: I would simplify a bit to just:
"Make sure you have installed Khiops properly."

Actually, the whole point of the "full Pip" installation route (alongside the "full Conda" one) is to abstract away the "Khiops Python library" / "Khiops binaries" dichotomy (which can be kept in the _developer_ comments though, but not in the user-facing warnings, messages or feedback IMHO).

khiops/core/internals/runner.py
```diff
@@ -987,7 +1000,8 @@ def _initialize_khiops_environment(self):
                 raise KhiopsEnvironmentError(
                     "The 'khiops_env' script not found for the current "
                     f"'{installation_method}' installation method. Make sure "
-                    "you have installed khiops >= 10.2.3. "
+                    "you have installed properly the Khiops Python library "
```
popescu-v: Simplify to "you have installed Khiops properly." (see relevant comment above).

khiops/core/internals/runner.py
```diff
@@ -1029,6 +1043,13 @@ def _initialize_khiops_environment(self):
                 elif var_name == "KHIOPS_MPI_COMMAND":
                     self._mpi_command_args = shlex.split(var_value)
                     os.environ["KHIOPS_MPI_COMMAND"] = var_value
+                # On Windows "KHIOPS_MPI_DLL_PATH" (containing the Intel MPI Library)
+                # must be added to "PATH" otherwise Khiops wouldn't find it
+                # and fail immediately
+                elif var_name == "KHIOPS_MPI_DLL_PATH":
```
popescu-v: This must only be done for Pip installations (whence _not_ for Conda). Cf. https://github.com/KhiopsML/khiops/blob/262e2c66788d63eb375c53e9301a283ac3cea82f/packaging/windows/khiops_env.cmd.in#L22.

khiops/core/internals/runner.py
```diff
@@ -1080,28 +1101,11 @@ def _initialize_khiops_version(self):
...
-                f"'{compatible_khiops_version.major}'. "
-                "To avoid any compatibility error, "
-                "please update either the Khiops "
-                f"executables package for your '{operating_system}' operating "
-                "system, or the Khiops Python library, "
-                f"according to your '{installation_method}' environment. "
-                "See https://khiops.org for more information.",
-            )
-
         # Warn if the khiops minor and patch versions do not match
```
popescu-v: I'm not sure all this still applies: if khiops-python is installed, then khiops-core is installed. If khiops-core is uninstalled by the user afterwards, then the environment initialization fails.

khiops/core/internals/runner.py
```diff
@@ -1080,28 +1101,11 @@ def _initialize_khiops_version(self):
 
         # Khiops core version
         self._khiops_version = KhiopsVersion(khiops_version_str)
-
         # Library version
         compatible_khiops_version = khiops.get_compatible_khiops_version()
-
         operating_system = platform.system()
         installation_method = _infer_khiops_installation_method()
```
popescu-v: I would remove all code after `self._khiops_version`, because the version mismatch would no longer be relevant here: it would be given by the package metadata (Conda or Pip) - see the comment below as well.

packaging/docker/khiopspydev/Dockerfile.ubuntu-debian
```diff
@@ -5,12 +5,12 @@ FROM ghcr.io/khiopsml/khiops/khiopsdev-${KHIOPSDEV_OS}:latest AS khiopsdev
 LABEL maintainer="khiops.team@orange.com"
 LABEL description="Container for the development of khiops-python"
 
-# Install dev tools and miniforge (for the unit tests); build and install Khiops
+# Install dev tools and miniforge (for the unit tests); install Khiops (only for tests against KhiopsDockerRunner)
```
popescu-v: I would drop this and set the `KHIOPS_CMD` and `KHIOPS_COCLUSTERING_CMD` to point to the specific Pip installation in the CI (see the other comments in this respect).

packaging/docker/khiopspydev/Dockerfile.ubuntu-debian
popescu-v: Drop this altogether and have the Docker runner use the Pip installation in the CI (see the other comments in this respect).

packaging/docker/khiopspydev/Dockerfile.ubuntu-debian
```diff
@@ -5,12 +5,12 @@ FROM ghcr.io/khiopsml/khiops/khiopsdev-${KHIOPSDEV_OS}:latest AS khiopsdev
...
 ARG KHIOPS_REVISION
```
popescu-v: Drop this and use virtualenv in the CI (see the other comments in this respect).

popescu-v (changes requested):
See pending comments.

khiops/core/internals/runner.py
tramora: At least during the making of the current PR, the case occurs :
- Khiops is 11.0.1rc1
- The Khiops Library is 11.0.0.3 (in the PR branch)
Surely the khiops team won't let this occur in the final release but who knows ?

khiops/core/internals/runner.py
tramora: I'm not very comfortable with the vague "Khiops" word as we show messages to the end user.

Indeed the user is using "The Khiops Python Library" so it is his/her starting point that why I would like to be precise on this point but I will follow

popescu-v: Ok, then we could keep "Make sure you have installed the Khiops Python library correctly.", without mentioning the "and Khiops" and the latter being installed differently, because it would be out of the scope of the "full Pip" installation IMHO.

On the other hand, if a user deliberately chooses to uninstall the `khiops-core` package in order to replace the binaries with, say, a locally-built Khiops Core set of binaries, then that user would not need this kind of warning in the first place, IMHO.

khiops/core/internals/filesystems.py
tramora: The blocking issue for the requesteed change is that `KhiopsLocalRunner` depends on `khiops.core.internals.filesystems`, not the other way. Changing this would let sub-optimal choices in my opinion :
- modifying the fs interface to pass a runner instance (for example in `create_resource`) which is a huge set of modifications and not elegant at all
- instanciating directly a runner inside fs that would introduce a circular dependency.
I'm thinking about the cleanest solution but we could totally remove this check also (even if we will lose a valuable feedback for the user)

khiops/core/internals/filesystems.py
tramora: Fixed

.github/instructions/docker-changes.instructions.md
tramora: I didn't think about reading the khiops-server source code. I implemented this solution to get rid of the native package

.github/workflows/dev-docker.yml
tramora: I didn't think about reading the khiops-server source code. I implemented this solution to get rid of the native package

khiops/core/internals/filesystems.py
tramora: Fixed

khiops/core/internals/runner.py

khiops/core/internals/runner.py

khiops/core/internals/runner.py
tramora: After all I left "Make sure you have installed Khiops properly." then.

khiops/core/internals/runner.py
tramora: After all I left "Make sure you have installed Khiops properly." then.

khiops/core/internals/runner.py
tramora: Added the "pip" installation_method condition

khiops/core/internals/runner.py
tramora: I got rid of the whole block regarding the minor version check

khiops/core/internals/runner.py
tramora: I got rid of the whole block regarding the minor version check

packaging/docker/khiopspydev/Dockerfile.ubuntu-debian
tramora: I didn't think about reading the khiops-server source code. I implemented this solution to get rid of the native package

packaging/docker/khiopspydev/Dockerfile.ubuntu-debian
tramora: I didn't think about reading the khiops-server source code. I implemented this solution to get rid of the native package

packaging/docker/khiopspydev/Dockerfile.ubuntu-debian
tramora: I didn't think about reading the khiops-server source code. I implemented this solution to get rid of the native package

.github/instructions/docker-changes.instructions.md
```diff
@@ -49,21 +45,16 @@ All Dockerfiles accept these `ARG` values, supplied by `dev-docker.yml`:
...
 
 ## Multi-Stage Build Structure
 
 Each Dockerfile uses a multi-stage build:
 
 1. **`khiopsdev`** — Based on `ghcr.io/khiopsml/khiops/khiopsdev-<os>:latest`.
-   Installs dev tools (git, pip, pandoc, wget), the Khiops native binary, and
-   Miniforge for Conda. On Debian/Ubuntu, also installs system-wide remote file
-   drivers (GCS, S3, Azure `.deb` packages) and `ruby-dev` (needed for fakeS3).
+   Installs dev tools (git, pip, pandoc, wget, zip, unzip), the Khiops native binary, and
```
popescu-v: Remove ", the Khiops native binary,".

packaging/docker/khiopspydev/Dockerfile.ubuntu-debian
```diff
@@ -0,0 +1,71 @@
...
+    done;' \
+  && true
+
+RUN mkdir -p /scripts
+COPY ./run_service.sh ./run_fake_remote_file_servers.sh /scripts/
+RUN chmod +x /scripts/run_service.sh /scripts/run_fake_remote_file_servers.sh && \
+  useradd -rm -d /home/ubuntu -s /bin/bash -g root -u 1000 ubuntu
+
+# The khiops-server will not use the Khiops binaries from the installation of Native packages
+# but the ones set up in the `KHIOPS_CMD` and `KHIOPS_COCLUSTERING_CMD` environment variables instead
```
popescu-v: I'm not sure this comment belongs here. I'd rather put it in the commit message of the commit that removes the native Khiops binaries installation lines.

khiops/core/internals/filesystems.py
popescu-v: Then, we could add `khiops_drivers_library_path` to the runner and do the check after each resource creation via `filesystems.create_resource` only `if not is_local_resource(created_resource)`. We can also factor out the calls to these last 2 functions into a public helper function inside `filesystems.py`, like `check_driver_library_path_for_resource(driver_library_dir, resource)`.

.github/workflows/pip.yml
```diff
@@ -77,6 +77,12 @@ jobs:
         run: |
           SAMPLES_REVISION=${{ inputs.samples-revision || env.DEFAULT_SAMPLES_REVISION }}
           echo "SAMPLES_REVISION=$SAMPLES_REVISION" >> "$GITHUB_ENV"
+      - name: Checkout sources  # Checkout the sources to be able to extract the dependencies list
+        uses: actions/checkout@v4
+        with:
+          # Get Git tags so that versioneer can function correctly
+          # See issue https://github.com/actions/checkout/issues/701
```
popescu-v: I would drop the mention of "versionneer", as it is no longer used. However, I'd keep the `fetch-depth` instruction, because the issue referred to in the comment is still open.

khiops/core/internals/runner.py
```diff
@@ -243,11 +243,12 @@ def _infer_khiops_installation_method(trace=False):
        it was not activated previously nor during the execution
        and thus the CONDA_PREFIX environment variable is undefined
        and the path to the `bin` directory inside the conda environment is not in PATH
-    - 'binary+pip' installs the binaries and the shared libraries system-wide
-      but will keep the python libraries
-      in the python system folder
-      or in the Python folder inside the home directory of the user,
-      or in a virtual environment (if one is used)
+    - 'pip' environment containing binaries, shared libraries and the Python libraries
+      can either be :
```
popescu-v: s/can either be :/can either be:/ (drop the space before the colon).

popescu-v (changes requested):
Just a few loose ends, among which a rather important one IMHO.

khiops/core/internals/filesystems.py
tramora: After a second look, the `filesystems` module was meant to be used in a very independent way (to confirm with @folmos-at-orange): 
`fs.write`, `fs.read` can be called at any time. This helper module helps manipulate and abstract any local or remote resource for testing purposes and it would be annoying if a tie with the runner is created.
I'd rather remove this new `_check_khiops_driver_library` and all the calls than risking to lose this independence

.github/instructions/docker-changes.instructions.md
tramora: Corrected

packaging/docker/khiopspydev/Dockerfile.ubuntu-debian
tramora: I moved this to the commit message for the unique commit that was already responsible for removing the old docker files and factor out debian and ubuntu OS, it would be a bit painful to create a new commit for this

.github/workflows/pip.yml
tramora: I removed the reference to the versioneer

khiops/core/internals/runner.py

khiops/core/internals/filesystems.py
tramora: We agreed to remove this extra-test

.github/instructions/docker-changes.instructions.md
popescu-v: I still see the mention of the "Khiops native binary". We don't install these native binaries anymore (because we use the Pip package), do we?

khiops/core/internals/runner.py
```diff
@@ -942,6 +943,7 @@ def __init__(self):
         self._mpi_command_args = None
         self._khiops_path = None
         self._khiops_coclustering_path = None
+        self._khiops_drivers_path = None
```
popescu-v: I would remove this "driver path" support code altogether, as it is not needed in the code base.

Indeed, the `KHIOPS_DRIVERS_PATH` environment variable is used by the Khiops binaries to locate the driver, which they need. The Khiops Python library is (and should remain IMHO) transparent to this, as it does not use the drivers itself, neither it performs any checks in this respect whatsoever (as we did not keep the driver library check in the Python library).

khiops/core/internals/runner.py
```diff
@@ -1428,6 +1408,28 @@ def khiops_coclustering_path(self, modl_coclustering_path):
...
+        if not os.path.exists(khiops_drivers_path):
+            raise KhiopsEnvironmentError(
+                f"Nonexistent Khiops drivers path: '{khiops_drivers_path}'"
+            )
+        if not os.path.isdir(khiops_drivers_path):
+            raise KhiopsEnvironmentError(
+                "Khiops drivers path is not a directory: " f"{khiops_drivers_path}"
+            )
+        self._khiops_drivers_path = khiops_drivers_path
+
```
popescu-v: I'd remove all this "drivers path" management code (see the comment above for the rationale of it).

popescu-v (changes requested):
Just a text to remove (mention of Khiops binaries in the Docker file), and some unused code to remove from the runner (see the pending comments).

.github/instructions/docker-changes.instructions.md
tramora: My bad, I was wrongly focusing on the punctuation. Fixed now

khiops/core/internals/runner.py
tramora: Removed

khiops/core/internals/runner.py
tramora: Removed

popescu-v (approved):
LGTM

PR #584, opened 2026-06-08
Document external data table paths in various contexts

### LINKED ISSUE #560: Specifying additional tables : Problem with reference table name
#### Description
Hello,
I went into a problem while deploying a model through the core API.
Note that this is not blocking, the end user simply has to be cautious. 
Yet it seems as some kind of bug (or put it politely "weak conformity to the specification").

To put it briefly : when specifying a deployment on a table that reference an external table, one has to explicitly add the "SNB_" prefix to the name of the reference table.
This is non uniform with the other secundary tables.
I elaborate below with an example.

_What, Why_

Say you are using the khiops python library (core module) to execute a learn task and then a deploy task.
The root table for analysis goes something like this :

Root Dictionary subscriptions(subs_Id) {
    Categorical subs_Id;
    Categorical offer_name;
    Categorical targetCol;
    Table(Subs_Payments) payments;
    Entity(Subs_Offer) offer [offer_name];
};

This main dictionary embeds reference to two tables : one secondary table that lists the payments related to each subscriptions and a reference table that provide generic info on the offer that can be subscribed.
This implies that the kdic files contains something along the following lines :

Root Dictionary Subs_Offer(offer_name) {
    Categorical offer_name;
    // various fields
};

Dictionary Subs_Payments(subs_Id) {
    Categorical subs_Id;
    // various fields
};

Ok. We're all set. 
Now the API call to trigger learning through the khiops core library should be something like this (focus on the additional tables) :
kh.train_predictor(
    subscriptions_kdic_path,
    "subscriptions",
     subscriptions_filepath,
    "targetCol",
    supervised_report_file_path,
    additional_data_tables={
                            "payments":subs_payments_filepath,
                            "/Subs_Offer": offers_with_payment_options_filepath
    },                        
    max_trees=0,
)

Notice that above regular secondary tables are pointed to through the variable names, whereas external reference tables are  pointed to through the slash + dictionary name (Subs_Offer and not offer).

Now here is the tricky part. When calling the API for deployment, we would expect the additional tables to have same names. Only the table to be deployed should be edited with the prefix "SNB_".
Alas, the reference table should also be edited with the prefix "SNB_".

So the current syntax is :
kh.deploy_model(
    model_kdic_path,
    "SNB_subscriptions",
    subscriptions_filepath, 
    scored_subs_path, # Fichier de sortie
    additional_data_tables={
                            "payments": subs_payments_filepath,
                            "/SNB_Subs_Offer": offers_with_payment_options_filepath
    }
) 

"/Subs_Offer" has turned into "/SNB_Subs_Offer".

Note that I have illustrated the problem with a small adhoc dico, inspired by the data I am currently working on.
But there is a dataset in the khiops data sample that might be used to work on this issue : it is the CustomerExtended dataset.
The main table Customer has a sub-table Address, that points to a reference table City.
Hence the spec of additional table should be something like : "/City" (when calling for train).
And when calling for deployment it becomes "/SNB_City".

#### Questions/Ideas
- _A potential solution__
Ok, so the template suggests that I write dow a potential solution.
Alas I am not sure about this one.
When specifying paths to the additional tables in the core api, I understand that external reference tables are actually refered to by their dictionary name. 
For instance, in the example I provided, the entity variable offer is actually of type Subs_Offer hence we put : "/Subs_Offer".
Then after training, in the kdic model file, all tables have been renamed with the "SNB_" prefix. There is no longer a Subs_Offer dictionary.

popescu-v:
- In `train_predictor`, `SNB_` or `Baseline_`-prefixed model dictionaries are generated. Similarly, in `train_recoder`, `R_`-prefixed model dictionaries are generated.

- In `deploy_model`, these prefixed dictionaries must be used, even for external tables.

- However, in `evaluate_predictor`, the original names must be used for external tables, in order to support evaluating several predictors.

_Put your message here_

---

### DIFF
--- a/doc/multi_table_primer.rst
+++ b/doc/multi_table_primer.rst
@@ -208,6 +208,28 @@ Types of secondary tables include:
     than the number of addresses so it may make sense to load it entirely in memory for efficiency
     reasons.
 
+  .. note:: In `.khiops.core.api.evaluate_predictor`, the initial name of the
+      external table must be used, which is the same as the one used for
+      training the predictor. 
+      For example:
+
+      .. code-block:: c
+
+         additional_data_tables: {
+             "/City": "/path/to/Cities.csv"
+         }
+
+  .. note:: In `.khiops.core.api.deploy_model`, model dictionary data paths must be used for
+       external tables. For example, when deploying a classification or a regression
+       model:
+
+       .. code-block:: c
+
+          additional_data_tables: {
+              "/SNB_City": "/path/to/Cities.csv"
+          }
+
+
 Note that besides the root table names the components of a data path are **table variable names**
 and not *table names*. For further details about the multi-table capabilities of Khiops refer to the
 documentation at `the Khiops site <https://khiops.org/setup/KhiopsGuide.pdf>`_.
--- a/khiops/core/api.py
+++ b/khiops/core/api.py
@@ -712,6 +712,11 @@ def train_predictor(
 ):
     r"""Trains a model from a data table
 
+    .. note::
+         For all input dictionaries, this function creates model dictionaries whose
+         names are prefixed with ``SNB_``. For regression models, additional
+         dictionaries are created and their names are prefixed with ``Baseline_``.
+
     Parameters
     ----------
     dictionary_file_path_or_domain : str or `.DictionaryDomain`
@@ -1069,7 +1074,9 @@ def evaluate_predictor(
         A dictionary containing the data paths and file paths for a multi-table
         dictionary file. For more details see :doc:`/multi_table_primer`.
 
-        .. note:: Use the initial dictionary name in the data paths.
+        .. note::
+             For external tables, use the initial dictionary name in the data paths,
+             which is the same as the one used for training the predictor.
 
     main_target_value : str, default ""
         If this target value is specified then it guarantees the calculation of lift
@@ -1164,6 +1171,10 @@ def train_recoder(
 
     The output files of this process contain a dictionary file (``.kdic``) that can be
     used to recode databases with the `deploy_model` function.
+
+    .. note::
+      For all input dictionaries, this function creates model dictionaries whose names
+      are prefixed with ``R_``.
 
     Parameters
     ----------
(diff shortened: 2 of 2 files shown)


doc/multi_table_primer.rst
```diff
@@ -208,6 +208,27 @@ Types of secondary tables include:
     than the number of addresses so it may make sense to load it entirely in memory for efficiency
     reasons.
 
+  .. note:: In `.khiops.core.api.evaluate_predictor`, the initial name of the
+      external table must be used, thus supporting the evaluation of several models.
```
marcboulle: Le commentaire `, thus supporting the evaluation of several models.` n'est pas adapté au cas de l'API python.
Depuis la GUI, on peut choisir un ou plusieurs prédicteurs à évaluer, mais ce n'est pas le cas de l'API, ou on spécifie le prédicteur à évaluer.

Soit supprimer, ce commentaire, soit le changer en quelque chose du type `, which is the same as the one used for training the predictor.`

khiops/core/api.py
```diff
@@ -1069,7 +1074,9 @@ def evaluate_predictor(
         A dictionary containing the data paths and file paths for a multi-table
         dictionary file. For more details see :doc:`/multi_table_primer`.
 
-        .. note:: Use the initial dictionary name in the data paths.
+        .. note::
+             Use the initial dictionary name in the data paths. This allows to
```
marcboulle: Le data path ne contient plus le dictionary name pour l'entité principale. Cela ne concerne que les tables externes.

Changer 
```
Use the initial dictionary name in the data paths. This allows to
evaluate several predictors simultaneously.
```
en 
```
For external tables, use the initial dictionary name in the data paths
, which is the same as the one used for training the predictor.
```

marcboulle (changes requested):
Petites changement suggérés.
A discuter si besoin (le comportement de la GUI et de l'API ne sont pas tout à fait identiques).

doc/multi_table_primer.rst
popescu-v: I would go for the second option, as it seems more precise to me.

popescu-v: Done. Second option applied.

khiops/core/api.py
popescu-v: Done.

marcboulle (approved):
LGTM

PR #611, opened 2026-08-18
Convert sphinx to zensical

### LINKED ISSUE #581: Convert Sphinx Documentation to MkDocs
#### Description
The goal of this issue is to convert the Sphinx documentation that is generated from the docstrings of the Khiops Python library source code to the MkDocs format.

This is needed to facilitate the integration with the main Khiops web site documentation generation process, thusly being able to seamlessly support several versions of it.

popescu-v:
Closes https://github.com/KhiopsML/khiops-python/issues/581.

---

  - [x] Check the docs build **without** warning: see the log of the API Docs workflow
  - [x] Check that your changes render well in HTML: download the API Docs artifact and open `index.html`


khiops/sklearn/dataset.py
```diff
@@ -1229,7 +1231,7 @@ def create_khiops_dictionary(self):
 
         Returns
         -------
-        `.Dictionary`:
+        [`Dictionary`][]:
```
tramora, line 1234: A bit surprised by the new syntax of link-like reference without any link in the second group of brackets

.github/workflows/api-docs.yml
```diff
@@ -1,119 +1,88 @@
...
+      - name: Install Khiops core
+        run: |
+          # Create the .venv virtual environment and activate it
+          uv venv
+          . .venv/bin/activate
+
+          # Building the Python API docs always executes khiops-python's own
+          # tutorial notebooks (create-doc -t -d), so a real Khiops core
+          # install is required unconditionally.
+          uv pip install khiops-core==${{ inputs.khiops-revision || env.DEFAULT_KHIOPS_REVISION }}
```
tramora, line 54: "uv pip install" is kept (instead of "uv add") to be as close as possible to pip right ? (no lock file etc...)

.github/workflows/api-docs.yml
```diff
@@ -1,119 +1,88 @@
...
+          uv pip install tomli
           # First, install all dependencies except khiops-core and khiops-drivers-*
-          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --exclude-khiops-family > requires-no-khiops.txt          
-          python -m pip install --user `cat requires-no-khiops.txt`
+          uv run scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --exclude-khiops-family > requires-no-khiops.txt
+          uv pip install `cat requires-no-khiops.txt`
           # khiops-core and khiops-drivers-* must always be installed from TestPyPI in order to avoid distorting usage statistics
-          python scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --khiops-family-only > requires-khiops.txt
-          python -m pip install --user --index-url https://test.pypi.org/simple `cat requires-khiops.txt`
+          uv run scripts/extract_dependencies_from_pyproject_toml.py -f "pyproject.toml" --khiops-family-only > requires-khiops.txt
```
tramora, line 64: As the venv is actived,
"uv run" should be equivalent to "python...", right ?

tramora (approved):
No blocking remarks so far. I just put minor comments that can be ignored.

.github/workflows/api-docs.yml
popescu-v: Yes AFAIU, but keeping "uv run" is more explicit IMHO.

.github/workflows/api-docs.yml
popescu-v: There is a uv.lock file. `uv add` manages dependencies in the repo's "pyproject.toml" file, adding the dependency to "[project.dependencies]" AFAIU. Whereas `uv pip install` just does what bare `pip install` does, but in the `.venv` virtual environment that is created via Uv. The former is out of scope here, because the khiops-python dependencies in "pyproject.toml" are different from the documentation build dependencies that are in "doc/util/requirements.txt". And khiops-core is just a documentation build dependency here, that is required in order to execute the tutorials. The latter keeps updating the current Uv virtual environment with the dependencies that are required to build the doc.

bruno-at-orange (approved):
LGTM

.github/workflows/api-docs.yml
```diff
@@ -1,119 +1,88 @@
...
-          kh-download-datasets --force-overwrite --version ${{ inputs.khiops-samples-revision || env.DEFAULT_KHIOPS_SAMPLES_REVISION }}
-          kh-status
+          uv pip install .
+          uv run kh-download-datasets --force-overwrite --version ${{ inputs.khiops-samples-revision || env.DEFAULT_KHIOPS_SAMPLES_REVISION }}
+
+          # Discard error as this is expected (Uv is not yet officially
+          # supported by khiops, but this should pose no issue in practice)
+          # TODO: Revert as soon as issue
+          # https://github.com/KhiopsML/khiops-python/issues/602 is tackled
+          uv run kh-status || true
```
tramora, line 78: The TODO is legit but with the installation under uv we don't get an Error but a Warning instead.
In this case kh-status still returns a 0 code.

bruno-at-orange (approved):
LGTM

khiops/sklearn/dataset.py
popescu-v: Yes, these are file-local links which are looked-up directly via the identifier present to the left. Similar to what Sphinx did with the dot-prefixed syntax.

.github/workflows/api-docs.yml
popescu-v: Well, I needed to put this because apparently an error is issued and the return code is not zero and the pipeline failed. For example, in this workflow run: https://github.com/KhiopsML/khiops-python/actions/runs/32134479837/job/95702547306#step:6:150.

The culprit is here: https://github.com/KhiopsML/khiops-python/blob/015e844d18aeefc5d309016095b743555d35c6bf/khiops/core/internals/runner.py#L1255

tramora: I first thought it was due to the warning but yes there is indeed an error above
```
Error: Khiops Python library installation path '/home/runner/work/khiops-python/khiops-python/khiops' does not match the current python environment ('/home/runner/work/khiops-python/khiops-python/.venv/bin/python3'). Go to https://khiops.org/ for instructions to re-install it (preferably in a virtual environment).
```
It seems that `uv run` does not activate its virtual env

popescu-v: > It seems that `uv run` does not activate its virtual env

Indeed it does not, whence its explicit activation in the khiops-doc CI.

PR #623, opened 2026-09-20
Set the `KHIOPS_PROC_NUMBER` env var for each training run using the value of `max_cores`

### LINKED ISSUE #585: Use KHIOPS_PROC_NUMBER  when max_cores is set
#### Description
On a 32 cores machine, when `max_cores` is set to 4, khiops launches 32 cores and uses only 4. That's because it modifies the scenario used to launch khiops. 
A best approach would be to set `KHIOPS_PROC_NUMBER` to 4, khiops will only launch 4 cores.

tramora:
- the allocated number of CPU cores can never be greater than the value of `max_cores`

Fixes #585 

---

- API Docs (only if there are changes in docstrings, markdown files or samples):
  - If there are any problems it is faster to iterate by [building locally the API Docs](doc/README.md#build-the-documentation)


khiops/core/internals/runner.py
```diff
@@ -335,18 +335,30 @@ def _get_current_library_installer():
...
 
     # Ensure HOME is always set for OpenMPI 5+
     # (using KHIOPS_MPI_HOME if it exists)
     if "HOME" not in khiops_env:
         khiops_env["HOME"] = khiops_env.get("KHIOPS_MPI_HOME", "")
+    if system_settings is not None and system_settings.max_cores is not None:
+        # An additional environment variable (local to this specific run)
+        # must also be set to avoid allocating all the available CPU cores.
+        # Thus, allocated CPU cores = max number of CPU cores used
+        khiops_env["KHIOPS_PROC_NUMBER"] = system_settings.max_cores
```
popescu-v: This is too late: `KHIOPS_PROC_NUMBER` needs to be set in a specific environment in which `khiops_env --env` is launched, because it is therein that `KHIOPS_MPI_COMMAND` is computed with respect to `KHIOPS_PROC_NUMBER` (basically, `-n $KHIOPS_PROC_NUMBER` is added to `mpiexec`). Otherwise, `khiops_process_args` still contains the `mpiexec` command without the `-n` option set according to `KHIOPS_PROC_NUMBER`: indeed, the MPI command is given by `KhiopsLocalRunner._mpi_command_args`, which is computed when `khiops_env` is read.

Hence, IMHO the way forward is to build and propagate a custom environment (with a `KHIOPS_PROC_NUMBER` set according to `max_cores`) to the `khiops_env --env` subprocess.

tests/test_khiops_integrations.py
```diff
@@ -303,6 +304,24 @@ def test_runner_environment_for_openmpi5(self):
         if initial_home:
             os.environ["HOME"] = initial_home
 
+    def test_runner_environment_max_cores_sets_proc_number_env(self):
```
popescu-v: The relevant test is to:
1. set `max_cores` on an API function;
2. call the function and capture its local runner's state;
3. on this state, check that `._mpi_command_args contains `-n ...` according to `max_cores`.

For 2., use the `wrapt`-based aproach (cf. `test_sklearn.py`, where the `wrap_with_parameter_trace` is used). On first thought, I'd go for defining a `wrap_with_runner_trace` helper, whereby I'd trace the runner used when running the API function, by:
- wrapping the `KhiopsRunner.run` method inside the `wrap_with_runner_trace` to trace `self` (`self` being the `KhiopsLocalRunner` instance).
- then, on that `self` trace, I'd check `self._mpi_command_args` for the existence of the `["-n", max_cores]` sublist.

popescu-v (changes requested):
See the comments.

Issue #622, opened 2026-09-18
Enable System Parameters Globally


popescu-v:
Problem Statement
===============
The `max_cores` Core API functions parameter allows to restrict the number of CPU cores per Core API function call (as per issue #585) and the `n_cores` does the same per Sklearn estimator instance (as per issue #605).

The same is true for the `memory_limit_mb` and `temp_dir`, at the Core API function level.

However, this constraint needs to be enforced at each function call or estimator instantiation and there is no officially-supported mechanism for enforcing it at a global level.

Details
=====
The goal of this issue is to provide a global mechanism for setting the maximum CPU core number, the maximum memory used, and the temporary directory. These settings must:
- take effect for the remainder of the current Python process;
- until they are globally disabled;
- unless they are overridden via the local settings (see issues #585 and, for the CPU core number only, #605).

Specifications
===========
A set of functions are added to the Core API, for:
- updating the `KHIOPS_PROC_NUMBER`, `KHIOPS_MEMORY_LIMIT` and `KHIOPS_TMP_DIR` environment variables:
   - the values of these variables in the environment the current Python process runs in are memorized;
   - the values of these variables are updated, if set to non-`None` values;
   - the values of these variables are reset to the values they had in the environment the current Python process runs in, at the time the latter was launched, if they are set to `None` (this includes the case the functions are called with no parameters; see below for the concrete signatures of these functions);
- getting the current values of these environment variables.

Function signatures and docstrings:

```python
# The functions are defined in the khiops.core.api module
def set_default_max_cores(max_cores=None):
    """Sets the maximum number of cores used by Khiops Core API function calls

    The maximum number of cores used by Khiops is set globally, in that all subsequent Khiops
    Core API calls use `max_cores` CPU cores at most, unless their own `max_cores`
    argument is passed, in which case it overrides the global setting.

    This function modifies the environment Khiops runs in, so that it uses at most `max_cores`
    CPU cores.

    Parameters
    ------------
    max_cores : int, optional
        Maximum number of CPU cores to be used by subsequent Khiops Core API function calls.
        If not set, then the function sets the maximum number of cores to the value it had prior to
        launching the Khiops Python process.
    """

def get_default_max_cores():
    """Gets the maximum number of cores used by Khiops Core API function calls"""

def set_default_memory_limit_mb(memory_limit_mb=None):
    """Sets the maximum amount of memory (megabytes) used by Khiops Core API function calls

[...]

PR #624, opened 2026-09-22
Enable System Parameters Globally #622

lucaurelien:


khiops/core/internals/runner.py
```diff
@@ -237,7 +239,7 @@ def _check_conda_env_bin_dir(conda_env_bin_dir):
     return is_conda_env_bin_dir
 
 
-def _infer_khiops_installation_method(trace=False):
+def _infer_khiops_installation_method(trace=False, environment=None):
```
popescu-v: To be consistent with the rest of the code base (to add to relevant instruction markdown?, put the `environment` kwarg before the `trace` kwarg.

doc/site/notes.md
```diff
@@ -61,6 +61,33 @@ The functions in the [khiops.core.api][] have the following common parameters.
...
+
+- [set_default_max_cores][khiops.core.api.set_default_max_cores] and
+    [get_default_max_cores][khiops.core.api.get_default_max_cores] configure and query
+    `KHIOPS_PROC_NUMBER`.
+- [set_default_memory_limit_mb][khiops.core.api.set_default_memory_limit_mb] and
+    [get_default_memory_limit_mb][khiops.core.api.get_default_memory_limit_mb]
+    configure and query `KHIOPS_MEMORY_LIMIT`.
+- [set_default_temp_dir][khiops.core.api.set_default_temp_dir] and
+    [get_default_temp_dir][khiops.core.api.get_default_temp_dir] configure and query
+    `KHIOPS_TMP_DIR`.
```
popescu-v: These 3 environment variables should not be documented explicitly for the users (they are an internal technical detail, specific to `khiops_env`. Hence, I'd replace the mention of these vars with "the maximum number of CPU cores", "the maximum amount of memory (in megabytes)", and "Khiops temporary directory", respectively.

doc/site/notes.md
```diff
@@ -61,6 +61,33 @@ The functions in the [khiops.core.api][] have the following common parameters.
...
+
+The numeric setters accept `None` or a positive `int`. The temporary-directory
+setter accepts `None` or a non-empty `str`; invalid types and values raise
```
popescu-v: `None` just means the parameter is optional. To observe coherence with the remainder of the API, I'd say "The numeric setters accept an optional positive `int`." Idem  for the temp dir variable. Also, I'd replace "Calling a setter with `None`" with "Calling a setter with no arguments".

doc/site/notes.md
```diff
@@ -61,6 +61,33 @@ The functions in the [khiops.core.api][] have the following common parameters.
...
+`TypeError` or `ValueError`, and a rejected update leaves the previous value
+unchanged. Calling a setter with `None` disables only that default and restores
+the corresponding value from the environment captured when the Khiops package
+initialized, or removes the variable if it was absent. Later changes to
+`os.environ` do not change this reset target. Getters return typed numeric
+values, a non-empty path, or `None` when the effective variable is absent.
```
popescu-v: Here `None` should be kept, because it's a return value the user must be aware of.

khiops/core/internals/runner.py
```diff
@@ -60,15 +60,17 @@ def get_default_samples_dir():
...
-        samples_dir = os.path.join(os.environ["PUBLIC"], "khiops_data", "samples")
+    environment = os.environ if environment is None else environment
+    if "KHIOPS_SAMPLES_DIR" in environment and environment["KHIOPS_SAMPLES_DIR"]:
+        samples_dir = environment["KHIOPS_SAMPLES_DIR"]
+    elif platform.system() == "Windows" and "PUBLIC" in environment:
+        samples_dir = os.path.join(environment["PUBLIC"], "khiops_data", "samples")
     else:
         # The filesystem abstract layer is used here
         # as the path can be either local or remote
+        home_dir = environment.get("HOME", environment.get("KHIOPS_MPI_HOME", ""))
```
popescu-v: No, this fallback to `KHIOPS_MPI_HOME` is only specific to OpenMPI >= 5 and is already being taken care of in `_build_khiops_process_environment`, which is called in `KhiopsLocalRunner.raw_run` just before launching the Khiops command (`mpiexec` included) in a subprocess, thus isolated to where it is needed.

This fallback is irrelevant to looking the samples directory, so I'd determine `samples_dir` as before, but:
- using `environment` rather than `os.environ`;
- in the scope of an `if "HOME" in environment` conditional block.

khiops/core/internals/runner.py
```diff
@@ -255,6 +257,8 @@ def _infer_khiops_installation_method(trace=False):
       - or in a classical virtual environment (highly encouraged)
 
     """
+    environment = os.environ if environment is None else environment
```
popescu-v: I'd factor this out in a small helper:

```python
def _current_environment(environment=None):
    if environment is not None:
        return environment
    return os.environ
```

Or in a more concise manner:

```python
def _current_environment(environment=None):
    return environment or os.environ
```

khiops/core/internals/runner.py
```diff
@@ -335,13 +339,16 @@ def _get_current_library_installer():
...
     """Build a specific environment used for the execution of khiops in a process
 
     This environment can be modified freely without interfering
     with the global one.
     """
-    khiops_env = os.environ.copy()
+    if environment is None:
+        khiops_env = os.environ.copy()
+    else:
+        khiops_env = environment.copy()
```
popescu-v: I'd invert the loop, to state that the `environment is None` is the fallback situation:

```python
if environment is not None:
    khiops_env = environment.copy()
else:
    khiops_env = os.environ.copy()

File: khiops/core/internals/runner.py [hunk 8]
```diff
@@ -946,8 +953,17 @@ class KhiopsLocalRunner(KhiopsRunner):
 
     """
 
-    def __init__(self):
+    def __init__(self, env=None):
```
popescu-v: I'd rename the kwarg to `environment` to be consistent with the other functions; or, otherwise, rename it to `env` everywhere.

khiops/core/internals/runner.py
```diff
@@ -962,7 +978,12 @@ def __init__(self):
         self._initialize_khiops_environment()
 
     def _initialize_khiops_environment(self):
-        installation_method = _infer_khiops_installation_method()
+        runner_environment = (
+            os.environ if self._environment is None else self._environment
```
popescu-v: Invert the loop to state that using `os.environ` is the fallback ("else") situation - see the previous comment in this respect.

khiops/core/internals/runner.py
```diff
@@ -1054,41 +1082,47 @@ def _initialize_khiops_environment(self):
...
-                    os.environ["KHIOPS_MPI_COMMAND"] = var_value
+                    runner_environment["KHIOPS_MPI_COMMAND"] = var_value
                 # On Windows, in 'pip' installations,
                 # "KHIOPS_MPI_DLL_PATH" (containing the Intel MPI Library)
                 # must be added to "PATH" otherwise Khiops wouldn't find it
                 # and fail immediately
                 elif installation_method == "pip" and var_name == "KHIOPS_MPI_DLL_PATH":
-                    os.environ["PATH"] = os.pathsep.join(
-                        [var_value, os.environ.get("PATH")]
+                    current_path = runner_environment.get("PATH")
```
popescu-v: I would have just kept the code as before and just replaced `os.environ` with `runner_environment`; `current_path` is not used elsewhere. Black should know how to format such line, even if it is longer than 88 characters.

khiops/core/internals/runner.py
```diff
@@ -1054,41 +1082,47 @@ def _initialize_khiops_environment(self):
...
+                    runner_environment["PATH"] = (
+                        os.pathsep.join([var_value, current_path])
+                        if current_path
+                        else var_value
                     )
+                    if self._environment is not None:
+                        runner_environment[var_name] = var_value
```
popescu-v: If we want to store `KHIOPS_MPI_DLL_PATH` in the environment in the case of the Pip installation method (for other installation methods, it is covered by the `else` below), I would just do, like for the other explicitly-addressed variables, just:
`runner_environment["KHIOPS_MPI_DLL_PATH"] = var_value`, whichever the `runner_environment` was (that is, we add this to `os.environ` as well, for consistency), hence the  `if self._environment is not None` disappears.

khiops/core/internals/runner.py
```diff
@@ -1054,41 +1082,47 @@ def _initialize_khiops_environment(self):
...
         # Check the tools exist and are executable
         self._check_tools()
 
         # Initialize the default samples dir
         self._initialize_default_samples_dir()
 
-    def _infer_khiops_env_from_path(self, installation_method):
-        khiops_env_path = shutil.which("khiops_env")
+    def _infer_khiops_env_from_path(self, installation_method, environment=None):
+        search_path = None if environment is None else environment.get("PATH", "")
```
popescu-v: Add comment before this line stating: "Fall back to searching `khiops_env` according to the `PATH` set in `os.environ`" and invert the `if`:
`search_path = environment.get("PATH") if environment is not None else None`

And do not fallback to "" if `PATH` is not defined in `environment` because this is not legit behavior as per `shutil.which` documentation (see https://docs.python.org/3/library/shutil.html#shutil.which): "path is a “PATH string” specifying the directories to look in, delimited by [os.pathsep](https://docs.python.org/3/library/os.html#os.pathsep). When no path is specified, the PATH environment variable is read from [os.environ](https://docs.python.org/3/library/os.html#os.environ), falling back to [os.defpath](https://docs.python.org/3/library/os.html#os.defpath) if it is not set."

khiops/core/internals/runner.py
```diff
@@ -1162,7 +1196,12 @@ def _detect_library_installation_incompatibilities(self, library_root_dir_path):
         error_list = []
         warning_list = []
 
-        installation_method = _infer_khiops_installation_method()
+        runner_environment = (
+            os.environ if self._environment is None else self._environment
```
popescu-v: Invert the if condition (cf. previous comments).

khiops/core/internals/runner.py
```diff
@@ -1335,7 +1374,12 @@ def _build_status_message(self):
         )
 
         # Build the messages for install type and mpi
-        install_type_msg = _infer_khiops_installation_method()
+        runner_environment = (
+            os.environ if self._environment is None else self._environment
```
popescu-v: Invert if to fall back on `os.environ` (cf. comments above).

khiops/core/api.py
```diff
@@ -29,9 +31,176 @@
...
+
+    if expected_type is int:
+        if isinstance(value, bool) or not isinstance(value, int):
+            raise TypeError(type_error_message(parameter_name, value, int))
+        if value <= 0:
+            raise ValueError(f"{parameter_name} must be positive (it is {value})")
+    elif not isinstance(value, str):
+        raise TypeError(type_error_message(parameter_name, value, str))
+    elif not value:
+        raise ValueError(f"{parameter_name} must be non-empty")
```
popescu-v: For consistency with the remainder of the code base and concision, I'd rewrite this snippet thusly:
```python
if not isinstance(value, expected_type):
    raise TypeError(type_error_message(parameter_name, value, expected_type))
if not value:
    raise ValueError(f"{parameter_name} must be non-empty")
if isinstance(value, int) and value < 0:
    raise ValueError(f"{parameter_name} must be positive; it is {value}")
```

khiops/core/api.py
```diff
@@ -17,7 +17,9 @@
 """
 import io
 import os
+import threading as _threading
```
popescu-v: Just `import threading` to stay consistent with the remainder of the code base. There is no need to make it "private" by convention via aliasing. Idem for `MappingProxyType`.

khiops/core/api.py
```diff
@@ -29,9 +31,176 @@
     is_string_like,
     type_error_message,
 )
-from khiops.core.internals.runner import get_runner
+from khiops.core.internals.runner import KhiopsLocalRunner as _KhiopsLocalRunner
```
popescu-v: Just `import KhiopsLocalRunner`. There is no need for "private"-by-convention aliasing.

khiops/core/api.py
```diff
@@ -29,9 +31,176 @@
...
+from khiops.core.internals.runner import get_runner, set_runner
 from khiops.core.internals.task import get_task_registry
 
+# Capture the process environment before any runner initialization can occur.
+_inherited_environment = _MappingProxyType(os.environ.copy())
+_current_environment = dict(_inherited_environment)
+_default_system_parameters_lock = _threading.RLock()
```
popescu-v: I would:
- use capital-lettered variables here (as per PEP8 for global variables), e.g. `_INHERITED_ENVIRONMENT`
- develop the comment a bit, like:
   ```python
    # Capture the process environment before any runner initialization can occur.
    # Avoid race condition on os.environ (via MappingProxyType) and
    # make sure updates to it are not propagated to the current environment
    # (via .copy()).
    ```

khiops/core/api.py
```diff
@@ -29,9 +31,176 @@
...
+    """Set one global system parameter and install its runner."""
+    global _current_environment
+
+    _validate_default_system_parameter(parameter_name, value, expected_type)
+    with _default_system_parameters_lock:
+        updated_environment = _current_environment.copy()
+        if value is None:
+            inherited_value = _inherited_environment.get(environment_variable)
+            if inherited_value is None:
+                updated_environment.pop(environment_variable, None)
```
popescu-v: I would have made this snippet more explicit:
```python
if inherited_value is None and environment_variable in updated_environment:
    del updated_environment[environment_variable]
```

khiops/core/api.py
```diff
@@ -29,9 +31,176 @@
...
+
+
+def set_default_max_cores(max_cores=None):
+    """Set the process-wide default maximum number of Khiops cores.
+
+    Parameters
+    ----------
+    max_cores : int, optional
+        Positive maximum number of cores. If `None`, restore the value of
+        `KHIOPS_PROC_NUMBER` captured when the Khiops package initialized.
```
popescu-v: Do not leak `KHIOPS_PROC_NUMBER` in the user-facing docstrings. Stick to the docstrings presented in the issue specification. Idem for the others.

tests/test_core.py
```diff
@@ -19,10 +19,12 @@
 from copy import copy
 from difflib import unified_diff
 from pathlib import Path
+from types import MappingProxyType
 from unittest import mock
 
 import khiops
 import khiops.core as kh
+import khiops.core.api as core_api
```
popescu-v: Replace with "from khiops.core import api" and use "api" instead of "core_api" henceforth.

tests/test_core.py
```diff
@@ -1020,6 +1022,138 @@ def mocked_raw_run(*_, **__):
...
+            (kh.set_default_temp_dir, 42, kh.get_default_temp_dir),
+        ]
+        expected_values = [2, 2, 128, 128, "/launch/tmp", "/launch/tmp"]
+
+        for (setter, invalid_value, getter), expected_value in zip(
+            invalid_updates, expected_values
+        ):
+            with self.subTest(invalid_value=invalid_value):
+                with self.assertRaises((TypeError, ValueError)):
+                    setter(invalid_value)
```
popescu-v: I'd also check that the exception messages are as expected (see e.g. in test_core.KhiopsCoreIOTests.test_std_stream_errors` for examples).

tests/test_khiops_integrations.py
```diff
@@ -346,6 +347,98 @@ def test_runner_environment_initialization(self):
...
+            mock_popen.call_args_list[0].kwargs["env"], isolated_environment
+        )
+        final_environment = mock_popen.call_args_list[1].kwargs["env"]
+        self.assertEqual(final_environment["KHIOPS_PROC_NUMBER"], "7")
+        self.assertEqual(final_environment["KHIOPS_MEMORY_LIMIT"], "256")
+        self.assertEqual(final_environment["KHIOPS_TMP_DIR"], "/isolated/tmp")
+        self.assertEqual(final_environment["KHIOPS_RETURNED"], "yes")
+        self.assertEqual(final_environment["HOME"], "/isolated/home")
+
+    def test_default_runner_does_not_export_mpi_dll_path(self):
```
popescu-v: No, because this was technical debt to be fixed in this PR. Hence, I'd remove this test.

popescu-v (changes requested):
See the comments.

khiops/core/internals/runner.py
```diff
@@ -1054,41 +1089,45 @@ def _initialize_khiops_environment(self):
...
-                    os.environ["PATH"] = os.pathsep.join(
-                        [var_value, os.environ.get("PATH")]
+                    runner_environment["PATH"] = os.pathsep.join(
+                        [var_value, runner_environment.get("PATH")]
                     )
+                    runner_environment[var_name] = var_value
                 # Propagate all the other environment variables to Khiops binaries
                 else:
-                    os.environ[var_name] = var_value
+                    runner_environment[var_name] = var_value
```
popescu-v: Use `KHIOPS_MPI_DLL_PATH` instead of `var_name` to stay consistent with the treatment of the other variables. This was already suggested in the previous-review comment https://github.com/KhiopsML/khiops-python/pull/624#discussion_r4081121493.

.gitignore
```diff
@@ -6,6 +6,13 @@
 # Mac/OSX
 .DS_Store
 
+# OpenSpec artifacts
+openspec/
+.github/agents/openspec.agent.md
+.github/prompts/opsx-*.prompt.md
+.github/skills/openspec-*/
+.github/workflows/copilot-setup-steps.yml
+
```
popescu-v: These files are not committed. Either commit them (I'm not sure we have to), or remove this from this PR, and, possibly, commit all these in a new, different PR, dedicated to OpenSpec integration.

popescu-v (changes requested):
A couple of comments.

tests/test_khiops_integrations.py
```diff
@@ -346,6 +347,63 @@ def test_runner_environment_initialization(self):
...
+        )
+        tool_path = os.environ.get("SHELL", "/bin/sh")
+
+        mock_popen = MagicMock()
+        mock_popen.return_value.__enter__.return_value.communicate.return_value = (
+            "KHIOPS_MPI_HOME /isolated/mpi\n"
+            f"KHIOPS_PATH {tool_path}\n"
+            f"KHIOPS_COCLUSTERING_PATH {tool_path}\n"
+            "KHIOPS_MPI_COMMAND mpiexec\n"
+            "KHIOPS_API_MODE false\n"
```
popescu-v: Leave this to `True` (always the case from Python).

tests/test_khiops_integrations.py
```diff
@@ -346,6 +347,63 @@ def test_runner_environment_initialization(self):
...
+            "KHIOPS_RETURNED yes\n",
```
popescu-v: Remove (does not exist).

